Issue #31, opened 2022-03-01
Simple Credential Issuance and Validation, Verification


decentralgabe:
Functionality to provide a schema, issuer, and to-be-holder and 'issue' a credential. The credential should be valid according to the schema, spec, and its signature should be verifiable.

PR #37, opened 2022-03-04
Credential Builder

decentralgabe:
#31


decentralgabe:
May add some more tests before merge. I'm happy that the tests pass against the spec's example.

decentralgabe:
Added tests for the builder

cryptosuite/jwssignaturesuite.go
```diff
@@ -56,40 +61,37 @@ func (j JWSSignatureSuite) RequiredContexts() []string {
...
-
+func (j JWSSignatureSuite) Sign(s Signer, p Provable) error {
 	// create proof before CVH
 	proof := j.createProof(s.KeyID())
 
 	// prepare proof options
 	contexts, err := GetContextsFromProvable(p)
 	if err != nil {
-		return nil, errors.Wrap(err, "could not get contexts from provable")
+		return errors.Wrap(err, "could not get contexts from provable")
```
mistermoe, line 71: 🎉 meaningful error messages

did/model_test.go
```diff
@@ -3,9 +3,10 @@
 package did
 
 import (
-	"encoding/json"
 	"testing"
 
+	"github.com/goccy/go-json"
```
mistermoe, line 8: solid choice! being able to pass Context.context into Marshal and Unmarshal is awesome

vc/credential.go
```diff
@@ -0,0 +1,236 @@
+package vc
```
mistermoe, line 1: I think it'd be helpful to include package level documentation ([reference](https://talks.golang.org/2013/bestpractices.slide#15)). Can also use the `doc.go` convention

mistermoe, line 1: would it make more sense if this file was called `builder.go` given that's what's in it?

mistermoe (approved):
Overall, looks great! left a few nits.

cryptosuite/jsonwebkey2020.go
```diff
@@ -414,7 +417,7 @@ func (v JSONWebKeyVerifier) Verify(message, signature []byte) error {
 	return err
 }
 
-func NewJSONWebKeyVerifier(key PublicKeyJWK) (*JSONWebKeyVerifier, error) {
+func NewJSONWebKeyVerifier(kid string, key PublicKeyJWK) (*JSONWebKeyVerifier, error) {
```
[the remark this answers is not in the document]
decentralgabe: for linked data things, there's an expected `verificationMethod` that indicates a key-reference in a DID, or another URI-resolvable document; this is only required for the signer. The verifier _technically_ does not need this; however, the JWK library we use requires the `kid` in the public key you construct to match the `kid` in the signed message you're verifying...which is why on this specific verifier it's a required parameter.

vc/credential.go
decentralgabe: opened https://github.com/TBD54566975/did-sdk/issues/40

decentralgabe: will rename

PR #61, opened 2022-03-19
Presentation Request and Submission Handling

decentralgabe:


credential/exchange/submission.go
```diff
@@ -0,0 +1,318 @@
...
+
+// processInputDescriptor runs the input evaluation algorithm described in the spec for a specific input descriptor
+// https://identity.foundation/presentation-exchange/#input-evaluation
+// TODO(gabe) consider normalization of claims before processing
+func processInputDescriptor(id InputDescriptor, claims []PresentationClaim) (*processedInputDescriptor, error) {
+	constraints := id.Constraints
+	fields := constraints.Fields
+	if !(constraints == nil || len(fields) == 0) {
+		return nil, fmt.Errorf("invalid input descriptor without constraints and/or fields: %s", id.ID)
+	}
```
mistermoe, line 188: Is this logic backwards? i'm reading it as: if constraints isn't nil or there are fields, throw an error saying there aren't constraints or fields

mistermoe: image

given that `constraints` isn't a required property, does it seem correct to return an error? or does it make more sense to return an empty `processedInputDescriptor`?

decentralgabe: is this correct?
```
if constraints != nil && len(fields) != 0 {
```

mistermoe: @decentralgabe i think it'd be:
```go
if (constraints == nil || len(fields) == 0) {
    return nil, fmt.Errorf("invalid input descriptor without constraints and/or fields: %s", id.ID)
}
```

which reads as: if constraints is nil or there are no fields in constraints then there's nothing to do so throw err.

could do it the way you suggested but then the rest of the logic in the function would be nested in that conditional

decentralgabe: d'oh. thanks

decentralgabe:
```
	constraints := id.Constraints
	if constraints == nil {
		return nil, fmt.Errorf("unable to process input descriptor without constraints")
	}
	fields := constraints.Fields
	if len(fields) != 0 {
		return nil, fmt.Errorf("unable to process input descriptor without fields: %s", id.ID)
	}
```

mistermoe: awesome. I was actually curious about if `constraints.Fields` would panic if `constraints` was nil, so i tried it in go playground and the program executed successfully so i decided not to comment about it. The way you have it now reads much cleaner IMO

PR #101, opened 2022-06-11
Replace packr with stdlib embed

### LINKED ISSUE #85: Switch Packr to Embed
We use [packr](https://github.com/gobuffalo/packr). They recommend using [Go Embed](https://pkg.go.dev/embed). Let's make the switch.

philippgille:
This PR replaces github.com/gobuffalo/packr/v2 with Go standard library's `embed`.

It closes #85.

Main changes:
- Rename directories with test data to `testdata`
    - following Go best practices (`testdata` directories have a special meaning in that the Go toolchain ignores files in it)
- Move test data to their respective packages
    - Packr allows embedding of files in directories outside of the current file tree, but `embed` doesn't (e.g. `../testdata` doesn't work)
- Move known schemas to a central place, `schema` package
    - Because of the mentioned limitation (`embed` doesn't allow to embed from directories outside the current file tree), and to avoid duplicating all schema files
    - All packages using the known schema files already import the `schema` package, so there's no additional imports required
 - Remove packr related targets and code from the magefile

Currently it exposes one new public API, `schema.GetKnownSchema`, so that the packages importing `schema` can get the files as well. But alternatively we can move the known schemas and that `GetKnownSchema` method into a new `internal/schema` or `internal/known_schemas` package. `internal` packages are treated special by Go, in that all packages in this repo can access that package, but it won't be part of the public API anymore. I think this makes sense, but wanted to keep it simple for now, to have a first basis for discussing the PR. Please let me know what you think about this.

:warning: Starting with the state of the `main` on commit 6e6b96eb3f0b24d7e3c3decb18c893c13d856a56, without any code changes, running `go mod tidy` with Go 1.17 changed the `go.mod` and `go.sum` files. I guess it was forgotten to be run recently? In any case I included that as first commit in this PR.

Tested: `mage cbd` succeeds :green_circle:


decentralgabe (approved):
perfect! thanks for taking this 😄

Issue #49, opened 2022-03-14
Support did:web


decentralgabe:
https://w3c-ccg.github.io/did-method-web/

singyiu:
Hi @decentralgabe 
would love to get a better understanding on the context of this issue.
i.e. 
Any did:web implementation we can take reference from, or we need to start from stretch with the specification ?
Does ssi-sdk need the full implementation of did:web, or just some selected features ?
Thanks

decentralgabe:
Here's what I'm thinking (but feel free to tell me there's another perspective):
- [create a valid did:web document](https://w3c-ccg.github.io/did-method-web/#create-register), assuming one has a domain (take in those arguments...)
- [resolve a did:web document](https://w3c-ccg.github.io/did-method-web/#read-resolve) given a domain and return a Go object that can be turned into the JSON DID data model
- I don't believe [optional paths](https://w3c-ccg.github.io/did-method-web/#optional-path-considerations) need to be implemented to start with

Here's a reference impl linked to in the spec: https://github.com/decentralized-identity/web-did-resolver

decentralgabe:
fixed in #116

PR #116, opened 2022-06-28
Singyiu/feat did web 01

singyiu:
For issue #49
 
Implementation for did:web
according to method specification https://w3c-ccg.github.io/did-method-web
(create and resolve methods implemented with optional path supported, but not update and deactivate)

### DIFF
--- a/did/web.go
+++ b/did/web.go
@@ -19,13 +19,13 @@ import (
 type DIDWeb string
 
 const (
-	DID_WEB_WELL_KNOWN_URL_PATH = ".well-known/"
-	DID_WEB_DID_DOC_NAME        = "did.json"
+	DIDWebWellKnownURLPath   = ".well-known/"
+	DIDWebDIDDocFilename        = "did.json"
 )
 
 // CreateDoc constructs a did:web DIDDocument from a specific key type and its corresponding public key
 // This method does not attempt to validate that the provided public key is of the specified key type
-// The returned DIDDoucment is expected further turned into a JSON file named did.json
+// The returned DIDDocument is expected further turned into a JSON file named did.json
 // and stored under the expected path of the target web domain
 // specification: https://w3c-ccg.github.io/did-method-web/#create-register
 func (did DIDWeb) CreateDoc(kt crypto.KeyType, publicKey []byte) (*DIDDocument, error) {
@@ -135,8 +135,7 @@ func (did DIDWeb) Resolve() (*DIDDocument, error) {
 		return nil, err
 	}
 	var doc DIDDocument
-	err = json.Unmarshal(docBytes, &doc)
-	if err != nil {
+	if err = json.Unmarshal(docBytes, &doc); err != nil {
 		return nil, err
 	}
 	if !did.IsValidDoc(doc) {
--- a/did/web.go
+++ b/did/web.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 
 	"github.com/TBD54566975/ssi-sdk/crypto"
+	"github.com/TBD54566975/ssi-sdk/cryptosuite"
+	"github.com/sirupsen/logrus"
 )
 
 // did:web method specification
@@ -19,9 +21,28 @@ import (
 type DIDWeb string
 
 const (
-	DIDWebWellKnownURLPath   = ".well-known/"
-	DIDWebDIDDocFilename        = "did.json"
+	DIDWebWellKnownURLPath = ".well-known/"
+	DIDWebDIDDocFilename   = "did.json"
+	DIDWebPrefix           = "did:web:"
 )
+
+// keyTypeToLDKeyType converts crypto.KeyType to cryptosuite.LDKeyType
+func keyTypeToLDKeyType(kt crypto.KeyType) (cryptosuite.LDKeyType, error) {
+	switch kt {
+	case crypto.Ed25519:
+		return Ed25519VerificationKey2018, nil
+	case crypto.X25519:
+		return X25519KeyAgreementKey2019, nil
+	case crypto.Secp256k1:
+		return EcdsaSecp256k1VerificationKey2019, nil
+	case crypto.P256, crypto.P384, crypto.P521, crypto.RSA:
+		return cryptosuite.JsonWebKey2020, nil
+	default:
+		err := fmt.Errorf("unsupported keyType: %+v", kt)
+		logrus.WithError(err).Errorf("keyType %+v failed to convert to LDKeyType", kt)
+		return "", err
+	}
+}
 
 // CreateDoc constructs a did:web DIDDocument from a specific key type and its corresponding public key
 // This method does not attempt to validate that the provided public key is of the specified key type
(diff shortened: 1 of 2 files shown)


did/web_test.go
```diff
@@ -0,0 +1,60 @@
...
+	assert.NoError(t, err)
+	assert.Equal(t, string(didWebBasic), doc.ID)
+}
+
+func TestDIDWebCreateDocFileBytes(t *testing.T) {
+	pk, _, err := crypto.GenerateEd25519Key()
+	assert.NoError(t, err)
+	docBytes, err := didWebBasic.CreateDocBytes(crypto.Ed25519, pk)
+	assert.NoError(t, err)
+	assert.NotNil(t, docBytes)
```
decentralgabe, line 59: nit: `assert.NotEmpty` makes sure that the value is neither nil nor an empty value

did/web_test.go
```diff
@@ -0,0 +1,60 @@
...
+	assert.NoError(t, err)
+	assert.Equal(t, "https://localhost:8443/.well-known/did.json", docUrl)
+	docUrl, err = didWebOptionalPath.GetDocURL()
+	assert.NoError(t, err)
+	assert.Equal(t, "https://example.com/user/alice/did.json", docUrl)
+}
+
+func TestDIDWebResolveDocBytes(t *testing.T) {
+	_, err := didWebToBeResolved.ResolveDocBytes()
+	assert.NoError(t, err)
```
decentralgabe, line 35: any assertions that can be made about the resolution result?

did/web.go
```diff
@@ -0,0 +1,146 @@
...
+func (did DIDWeb) Resolve() (*DIDDocument, error) {
+	docBytes, err := did.ResolveDocBytes()
+	if err != nil {
+		return nil, err
+	}
+	var doc DIDDocument
+	err = json.Unmarshal(docBytes, &doc)
+	if err != nil {
+		return nil, err
+	}
```
decentralgabe, line 141:
```suggestion
	if  err = json.Unmarshal(docBytes, &doc); err != nil {
		return nil, err
	}
```

did/web.go
```diff
@@ -0,0 +1,146 @@
...
+// did:web method specification
+// https://w3c-ccg.github.io/did-method-web/
+// DID Web create and resolve mothods are implemented in this package
+// but NOT the update and deactivate methods
+// please refer to web_test.go for example and test cases
+type DIDWeb string
+
+const (
+	DID_WEB_WELL_KNOWN_URL_PATH = ".well-known/"
+	DID_WEB_DID_DOC_NAME        = "did.json"
```
decentralgabe, line 23:
```suggestion
	DIDWebWellKnownURLPath   = ".well-known/"
	DIDWebDIDDocFilename        = "did.json"
```

did/web.go
```diff
@@ -0,0 +1,146 @@
...
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/TBD54566975/ssi-sdk/crypto"
+)
+
+// did:web method specification
+// https://w3c-ccg.github.io/did-method-web/
+// DID Web create and resolve mothods are implemented in this package
```
decentralgabe, line 16:
```suggestion
// DID Web create and resolve methods are implemented in this package
```

did/web.go
```diff
@@ -0,0 +1,146 @@
...
+	DID_WEB_DID_DOC_NAME        = "did.json"
+)
+
+// CreateDoc constructs a did:web DIDDocument from a specific key type and its corresponding public key
+// This method does not attempt to validate that the provided public key is of the specified key type
+// The returned DIDDoucment is expected further turned into a JSON file named did.json
+// and stored under the expected path of the target web domain
+// specification: https://w3c-ccg.github.io/did-method-web/#create-register
+func (did DIDWeb) CreateDoc(kt crypto.KeyType, publicKey []byte) (*DIDDocument, error) {
+	//create DIDKey with publicKey
```
decentralgabe, line 32: nit: space before comment please
`// create DIDKey ... `

did/web.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	//create DIDDocument
+	didDoc, err := didKey.Expand()
+	if err != nil {
+		return nil, err
+	}
+	didWebStr := string(did)
+	didDoc.ID = didWebStr
+	verMethodID := didWebStr + "#owner"
+	didDoc.VerificationMethod[0].ID = verMethodID
```
decentralgabe, line 45: I'm not certain how I feel about relying so heavily on the did key methods. They have a separate API, and could change. Changes in the did:key API **should not** break the `did:web` methods.

What do you think about re-implementing some of the DID creation logic, specific to this DID method?

did/web.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	return json.Marshal(doc)
+}
+
+// GetDocURL returns the expected URL of the DID Document
+// where https:// prefix is required by the specification
+// optional path supported
+func (did DIDWeb) GetDocURL() (string, error) {
+	//DIDWeb must be prefixed with did:web:
+	if !strings.HasPrefix(string(did), "did:web:") {
```
decentralgabe, line 67: please make a constant
`const	DIDWebPrefix = "did:web" `

did/web.go
```diff
@@ -0,0 +1,146 @@
...
+		return "", fmt.Errorf("%+v not a valid did:web", did)
+	}
+
+	subStrs := strings.Split(string(did), ":")
+	numSubStrs := len(subStrs)
+	if numSubStrs < 3 || subStrs[0] != "did" || subStrs[1] != "web" {
```
decentralgabe, line 73: this seems like a duplicative check, considering the one above.

did/web.go
```diff
@@ -0,0 +1,146 @@
...
+		return "", fmt.Errorf("%+v not a valid did:web", did)
+	}
+	decodedDomain, err := url.QueryUnescape(subStrs[2])
+	if err != nil {
+		return "", err
+	}
+
+	//with well-known path
+	if numSubStrs == 3 {
```
decentralgabe, line 82: consider annotating this method with comments describing the steps as outlined here https://w3c-ccg.github.io/did-method-web/#read-resolve

e.g., this would be
```
// 3. Generate an HTTPS URL to the expected location of the DID document by prepending https://.
```

did/web.go
```diff
@@ -0,0 +1,146 @@
...
+		urlStr := "https://" + decodedDomain + "/" + DID_WEB_WELL_KNOWN_URL_PATH + DID_WEB_DID_DOC_NAME
```
decentralgabe, line 83: this is step 4
```
If no path has been specified in the URL, append /.well-known.
```

however, we need to check whether a path has been specified in the URL before appending the known path

did/web.go
```diff
@@ -0,0 +1,146 @@
...
+// on the expected URL of the DID Document from GetDocURL
+// and returns the bytes of the fetched file
+func (did DIDWeb) ResolveDocBytes() ([]byte, error) {
+	docUrl, err := did.GetDocURL()
+	if err != nil {
+		return nil, err
+	}
+	resp, err := http.Get(docUrl)
+	if err != nil {
+		return nil, err
```
decentralgabe, line 111: consider:
```logrus.WithError(err).Errorf("could not resolve did:web doc URL: %s", docURL)```

decentralgabe (changes requested):
please see changes. thanks for the work

did/web.go
singyiu: Good call. Will do.

did/web.go
singyiu: Will remove the duplicated condition. Split on ":" is on purpose such that it does not need to split again or replace string for supporting optional path.

singyiu:
@decentralgabe 
Thanks a lot for the review.
Various touch up had been made accordingly.

decentralgabe (approved):
nice work! thanks for making the changes.

Issue #114, opened 2022-06-27
Support did:peer


andorsk:
Support Peer DID: https://identity.foundation/peer-did-method-spec/

andorsk:
I can take this one @decentralgabe . DID peer is something I'm particularly interested in.

decentralgabe:
Fixed in #120

PR #120, opened 2022-06-30
feat: did:peer

andorsk:
This is an implementation of https://identity.foundation/peer-did-method-spec/ related to 

https://github.com/TBD54566975/ssi-sdk/issues/114. 

Implemented:

1. limited implementation of method 0 and method 2 or did:peer

Also included: 

1. DID and Resolver interfaces
2. Some shared stuff was refactored such as general key encoding/ errors
3. Update to the service struct and added some additional support to the DIDDocuments.

There's more to do here, but I will leave that for another PR unless it is necessary to add it to this one. Anything unimplemented I have returning an "unimplemented" error.

### DIFF
--- a/did/interface.go
+++ b/did/interface.go
@@ -1,0 +1,28 @@
+package did
+
+import (
+	"strings"
+
+	"github.com/TBD54566975/ssi-sdk/util"
+	"github.com/pkg/errors"
+)
+
+// TODO: All DID's MUST do X?
+// Define the interface more clearly
+type DID interface {
+	IsValid() bool
+	Resolve() (*DIDDocument, error)
+	ToString() string
+}
+
+type DIDURL interface {
+	Dereference() interface{}
+}
+
+func ParseDID(did DID, prefix string) (string, error) {
+	split := strings.Split(did.ToString(), prefix)
+	if len(split) != 2 {
+		return "", errors.Wrap(util.INVALID_FORMAT_ERROR, "did can't split correctly")
+	}
+	return split[1], nil
+}
--- a/did/interface.go
+++ b/did/interface.go
@@ -22,7 +22,7 @@ type DIDURL interface {
 func ParseDID(did DID, prefix string) (string, error) {
 	split := strings.Split(did.ToString(), prefix)
 	if len(split) != 2 {
-		return "", errors.Wrap(util.INVALID_FORMAT_ERROR, "did can't split correctly")
+		return "", errors.Wrap(util.InvalidFormatError, "did can't split correctly")
 	}
 	return split[1], nil
 }
--- a/did/interface.go
+++ b/did/interface.go
@@ -11,8 +11,8 @@ import (
 // Define the interface more clearly
 type DID interface {
 	IsValid() bool
-	Resolve() (*DIDDocument, error)
 	ToString() string
+	Parse() (string, error)
 }
 
 type DIDURL interface {
@@ -25,4 +25,8 @@ func ParseDID(did DID, prefix string) (string, error) {
 		return "", errors.Wrap(util.INVALID_FORMAT_ERROR, "did can't split correctly")
 	}
 	return split[1], nil
+}
+
+type Resolver interface {
+	Resolve(d DID) (*DIDDocument, error)
 }
--- a/did/model.go
+++ b/did/model.go
@@ -54,8 +54,8 @@ type Service struct {
 	// A string, map, or set composed of one or more strings and/or maps
 	// All string values must be valid URIs
 	ServiceEndpoint interface{} `json:"serviceEndpoint" validate:"required"`
-	RoutingKeys     []string    `json:"routingKeys"`
-	Accept          []string    `json:"accept"`
+	RoutingKeys     []string    `json:"routingKeys,omitempty"`
+	Accept          []string    `json:"accept,omitempty"`
 }
 
 func (s *Service) IsValid() bool {
--- a/did/model.go
+++ b/did/model.go
@@ -54,6 +54,8 @@ type Service struct {
 	// A string, map, or set composed of one or more strings and/or maps
 	// All string values must be valid URIs
 	ServiceEndpoint interface{} `json:"serviceEndpoint" validate:"required"`
+	RoutingKeys     []string    `json:"routingKeys"`
+	Accept          []string    `json:"accept"`
 }
 
 func (d *DIDDocument) IsEmpty() bool {
@@ -65,6 +67,17 @@ func (d *DIDDocument) IsEmpty() bool {
 
 func (d *DIDDocument) IsValid() error {
 	return util.NewValidator().Struct(d)
+}
+
+func NewDIDDocument() *DIDDocument {
+	return &DIDDocument{
+		Authentication:       make([]VerificationMethodSet, 0),
+		AssertionMethod:      make([]VerificationMethodSet, 0),
+		KeyAgreement:         make([]VerificationMethodSet, 0),
+		CapabilityDelegation: make([]VerificationMethodSet, 0),
+		CapabilityInvocation: make([]VerificationMethodSet, 0),
+		Services:             make([]Service, 0),
+	}
 }
 
 // TODO(gabe) DID Resolution Metadata
--- a/did/model.go
+++ b/did/model.go
@@ -58,6 +58,10 @@ type Service struct {
 	Accept          []string    `json:"accept"`
 }
 
+func (s *Service) IsValid() bool {
+	return !(s.ID == "" || s.Type == "" || s.ServiceEndpoint == "")
+}
+
 func (d *DIDDocument) IsEmpty() bool {
 	if d == nil {
 		return true
--- a/did/model.go
+++ b/did/model.go
@@ -59,7 +59,7 @@ type Service struct {
 }
 
 func (s *Service) IsValid() bool {
-	return !(s.ID == "" || s.Type == "" || s.ServiceEndpoint == "")
+	return util.NewValidator().Struct(s) == nil
 }
 
 func (d *DIDDocument) IsEmpty() bool {
--- a/did/peer.go
+++ b/did/peer.go
@@ -182,6 +182,12 @@ func (d DIDPeer) IsValidPurpose(p PurposeType) bool {
 	return false
 }
 
+// Resolves a did:peer into a DID Document
+// To do so, it decodes the key, constructs a verification
+// method, and returns a DID Document
+//
+// This allows PeerMethod0 to implement the DID Resolver
+// interface and be used to expand the did into the DID Document.
 func (m PeerMethod0) Resolve(did DID) (*DIDDocument, error) {
 
 	d, ok := did.(DIDPeer)
--- a/did/peer.go
+++ b/did/peer.go
@@ -25,10 +25,13 @@ import (
 	"github.com/TBD54566975/ssi-sdk/crypto"
 	"github.com/TBD54566975/ssi-sdk/util"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
-type DIDPeer string
-type PurposeType string
+type (
+	DIDPeer     string
+	PurposeType string
+)
 
 var availablePeerMethods = map[string]Resolver{
 	"0": PeerMethod0{},
@@ -55,7 +58,8 @@ const (
 func isPeerDID(did string) bool {
 	r, err := regexp.Compile(PeerDIDRegex)
 	if err != nil {
-		panic(err)
+		logrus.WithError(err).Error() // this should never happen
+		return false
 	}
 	return r.MatchString(did)
 }
@@ -68,13 +72,12 @@ func (d DIDPeer) IsValid() bool {
 
 func (d DIDPeer) Parse() (string, error) {
 
-	s, err := ParseDID(d, DIDPrefix+":"+PeerMethodPrefix+":")
-
+	s, err := ParseDID(d, strings.Join([]string{DIDPrefix, PeerMethodPrefix}, ":")+":")
 	if err != nil {
 		return "", err
 	}
 
-	method, err := d.GetMethodId()
+	method, err := d.GetMethodID()
 	if err != nil {
 		return "", err
 	}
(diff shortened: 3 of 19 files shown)


andorsk:
Things aren't nearly done here yet, but a couple key things I've realized in this process that I wanted to share:

1. I've added in an DID interface. I'm pretty sure we'll need this, but I need to align this with the DID specifications on DIF. 
2. Errors should be standardized. Probably in the "errors" file of the "utils", which is where I put things. 

I'll update this when I've made more progress.

did/interface.go
```diff
@@ -0,0 +1,23 @@
+package did
+
+import "strings"
+
+// TODO: All DID's MUST do X?
+// Define the interface more clearly
+type DID interface {
+	IsValid() bool
+	Resolve() (*DIDDocument, error)
```
decentralgabe, line 9: hmm, not sure about this one. 
some DIDs (like ION, and others) require network resolvers, which require hosted software. I'd like to avoid making recommendations of which resolvers to use here.

some DIDs (like peer, web, key) can be resolved more simply.

decentralgabe:
looks great so far. I like where your head is at for (1)..left a comment
(2)- agreed

did/interface.go
andorsk: Sure. I'm not sure about this one either. I do think we need a DID interface though! I'm just not sure exactly what goes into it. IsValid() seems reasonable. What else is required in a DID?

andorsk:
thanks @decentralgabe ! yep. Got them. Thanks!

did/interface.go
decentralgabe: hmm...maybe some common methods to get keys? like `getAuthenticationMethods`, `getVerificationMethods` and so on?

would validation just be static, or should we eventually expand to determining whether the keys in the doc are themselves valid?

there also needs to be a more flexible way to determine validity across DID methods, as they each have a set of properties/processes that uniquely apply to each method

andorsk: I think that's why each DID needs to implement an IsValid method no? That method would determine if the DID is valid or not. 

It seems like that the biggest issue here is the network resolvers?

andorsk: For resolution, here's what I ended up doing which works: 

```
type Resolver interface {
	Resolve(d DID) (*DIDDocument, error)
}
```

Some did's will have resolvers built in, and so will not. So network based resolvers we can ignore fine, it just won't be implementing the Resolver interface

andorsk:
Right now there is simple encoding/decoding of method0 and method2 of did:peer. I'd prefer to get this into `main`, and then submit subsequent PR's with additional support.

did/model.go
```diff
@@ -54,6 +54,12 @@ type Service struct {
 	// A string, map, or set composed of one or more strings and/or maps
 	// All string values must be valid URIs
 	ServiceEndpoint interface{} `json:"serviceEndpoint" validate:"required"`
+	RoutingKeys     []string    `json:"routingKeys"`
+	Accept          []string    `json:"accept"`
```
decentralgabe, line 58:
```suggestion
	RoutingKeys     []string    `json:"routingKeys,omitempty"`
	Accept          []string    `json:"accept,omitempty"`
```

did/model.go
```diff
@@ -67,4 +73,15 @@ func (d *DIDDocument) IsValid() error {
 	return util.NewValidator().Struct(d)
 }
 
+func NewDIDDocument() *DIDDocument {
+	return &DIDDocument{
+		Authentication:       make([]VerificationMethodSet, 0),
```
decentralgabe, line 78: would `new()` work? may be easier to read

did/peer_test.go
```diff
@@ -0,0 +1,184 @@
...
+		},
+		CapabilityDelegation: []VerificationMethodSet{
+			[]string{"#z6Mku9kBcbbGgp5G2oSPTqVsAqWhtTsNyPoxGvCRuQP9xDs6"},
+		},
+	}
+}
+
+func TestPeerMethod0(t *testing.T) {
+
+	var kt = crypto.Ed25519
```
decentralgabe, line 44:
```suggestion
	kt := crypto.Ed25519
```

did/peer_test.go
```diff
@@ -0,0 +1,184 @@
...
+	tDoc := makeSamplePeerDIDDocument1()
+
+	assert.Equal(t, tDoc.Context, doc.Context)
+
+}
+
+func TestPeerMethod2(t *testing.T) {
+
+	var d DIDPeer
+	var kt = crypto.Ed25519
```
decentralgabe, line 65:
```suggestion
	kt := crypto.Ed25519
```

did/peer_test.go
```diff
@@ -0,0 +1,184 @@
...
+// 	"routingKeys": ["did:example:somemediator#somekey"],
+//           "accept": ["didcomm/v2", "didcomm/aip2;env=rfc587"]
+// }
+// Service Block, after whitespace removal and common word substitution:
+// {"t":"dm","s":"https://example.com/endpoint","r":["did:example:somemediator#somekey"],"a":["didcomm/v2","didcomm/aip2;env=rfc587"]}
+// Encoded Service Endpoint: .SeyJ0IjoiZG0iLCJzIjoiaHR0cHM6Ly9leGFtcGxlLmNvbS9lbmRwb2ludCIsInIiOlsiZGlkOmV4YW1wbGU6c29tZW1lZGlhdG9yI3NvbWVrZXkiXSwiYSI6WyJkaWRjb21tL3YyIiwiZGlkY29tbS9haXAyO2Vudj1yZmM1ODciXX0=
+// Method 2 peer DID: did:peer:2.Ez6MkpTHR8VNsBxYAAWHut2Geadd9jSwuBV8xRoAnwWsdvktH.VzXwpBnMdCm1cLmKuzgESn29nqnonp1ioqrQMRHNsmjMyppzx8xB2pv7cw8q1PdDacSrdWE3dtB9f7Nxk886mdzNFoPtY.SeyJ0IjoiZG0iLCJzIjoiaHR0cHM6Ly9leGFtcGxlLmNvbS9lbmRwb2ludCIsInIiOlsiZGlkOmV4YW1wbGU6c29tZW1lZGlhdG9yI3NvbWVrZXkiXSwiYSI6WyJkaWRjb21tL3YyIiwiZGlkY29tbS9haXAyO2Vudj1yZmM1ODciXX0=
+func TestPeerResolveMethod2(t *testing.T) {
+
+	var tDoc = makeSamplePeerDIDDocument()
```
decentralgabe, line 153:
```suggestion
	tDoc := makeSamplePeerDIDDocument()
```

maybe something more descriptive, like `testDoc`?

util/errors.go
```diff
@@ -23,3 +23,12 @@ func LoggingErrorMsg(err error, msg string) error {
...
 	return errors.Wrap(err, msg)
 }
+
+// Error messages
+var (
+	UNSUPPORTED_ERROR     = errors.New("not supported")
+	NOT_IMPLEMENTED_ERROR = errors.New("not implemented")
+	INVALID_FORMAT_ERROR  = errors.New("invalid format")
+	UNDEFINED_ERROR       = errors.New("undefined")
+	CASTING_ERROR         = errors.New("failed to convert")
```
decentralgabe, line 33:
```suggestion
	UnsupportedError      = errors.New("not supported")
	NotImplementedError = errors.New("not implemented")
	InvalidFormatError  = errors.New("invalid format")
	UndefinedError       = errors.New("undefined")
	CastingError         = errors.New("failed to convert")
```

andorsk:
@decentralgabe updates made. One question for you and then maybe we are good to go?

did/peer.go
```diff
@@ -0,0 +1,539 @@
...
+// Peer based, self signed DID method.
+//
+// The method can be used independent of any central source of truth, and is
+// intended to be cheap, fast, scalable, and secure. It is suitable for most
+// private relationships between people, organizations, and things. We expect
+// that peer-to-peer relationships in every blockchain ecosystem can benefit by
+// offloading pairwise and n-wise relationships to peer DIDs.
+//
+// Currently only methods 0 and 2 are supported.
+// Method 1 will be supported in a future date
```
decentralgabe, line 14: created https://github.com/TBD54566975/ssi-sdk/issues/137

did/peer.go
```diff
@@ -0,0 +1,539 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/TBD54566975/ssi-sdk/crypto"
+	"github.com/TBD54566975/ssi-sdk/util"
+	"github.com/pkg/errors"
+)
+
+type DIDPeer string
+type PurposeType string
```
decentralgabe, line 31:
```suggestion
type (
    DIDPeer string
    PurposeType string
)
```

did/peer.go
```diff
@@ -0,0 +1,539 @@
...
+}
+
+//https://identity.foundation/peer-did-method-spec/#backing-storage
+type PeerDelta struct {
+	Change string    `json:"change"` // <base64url encoding of a change fragment>,
+	By     []byValue `json:"by"`     //  [ {"key": <id of key>, "sig": <signature value>} ... ],
+	When   int64     `json:"when"`   //<ISO8601/RFC3339 UTC timestamp with at least second precision>
+}
+
+func (a DIDPeer) Delta(b DIDPeer) (PeerDelta, error) {
```
decentralgabe, line 120: nit: keep receivers consistent `(d DIDPeer)`

did/peer.go
```diff
@@ -0,0 +1,539 @@
...
+	return PeerDelta{}, errors.Wrap(util.NotImplementedError, "peer:did delta")
```
decentralgabe, line 121: what do you think about returning `(*PeerDelta, error)`? I usually like pointer returns to simply return `nil, err` instead of instantiating an empty instance of the struct

did/peer.go
```diff
@@ -0,0 +1,539 @@
...
+}
+
+// TODO: CRDT
```
decentralgabe, line 124: is there an issue to track this? if not, please create one if CRDTs are a requirement for did:peer

did/peer.go
```diff
@@ -0,0 +1,539 @@
...
+}
+
+func (d DIDPeer) IsValidPurpose(p PurposeType) bool {
+	if _, ok := supportedDIDPeerPurposes[p]; ok {
+		return true
+	}
+	return false
+}
+
+func (m PeerMethod0) Resolve(did DID) (*DIDDocument, error) {
```
decentralgabe, line 182: comments here could help. it's not clear why these steps are required without reading through the spec

did/peer.go
```diff
@@ -0,0 +1,539 @@
...
+		ID:                   id,
+		VerificationMethod:   []VerificationMethod{*verificationMethod},
+		Authentication:       verificationMethodSet,
+		AssertionMethod:      verificationMethodSet,
+		KeyAgreement:         verificationMethodSet,
+		CapabilityDelegation: verificationMethodSet,
+	}, nil
+}
+
+func (m PeerMethod1) Resolve(d DID) (*DIDDocument, error) {
```
decentralgabe, line 222: ah I see here it's marked as not implemented. seems useful for handling method 1 dids we may encounter

did/peer.go
```diff
@@ -0,0 +1,539 @@
...
+
+// ANBF specified here:
+// https://identity.foundation/peer-did-method-spec/#method-specific-identifier
+const (
+	PeerMethodPrefix                = "peer"
+	DIDPrefix                       = "did"
+	PeerTransform                   = "z"
+	PeerEncNumBasis                 = Base58BTCMultiBase
+	PeerDIDRegex                    = `^did:peer:(([01](z)([1-9a-km-zA-HJ-NP-Z]{46,47}))|(2((\.[AEVID](z)([1-9a-km-zA-HJ-NP-Z]{46,47}))+(\.(S)[0-9a-zA-Z=]*)?)))$`
+	PeerKnownContext                = "https://w3id.org/did/v1"
```
decentralgabe, line 47: this is a duplicate of https://github.com/TBD54566975/ssi-sdk/blob/main/did/model.go#L12

did/peer.go
```diff
@@ -0,0 +1,539 @@
...
+		serviceBlock.Type = PeerDIDCommMessaging
+	}
+
+	return &serviceBlock, nil
+}
+
+type ServiceTypeAbbreviationMap map[string]string
+
+func (m PeerMethod2) Encode() ([]byte, error) {
+	return nil, util.NotImplementedError
```
decentralgabe, line 475: should there be a follow-up issue for this?

did/peer.go
```diff
@@ -0,0 +1,539 @@
...
+// Calculate the SHA256 [RFC4634] hash of the bytes of the stored variant of the
+// genesis version of the DID doc, and make this value the new DID's numeric
+// basis.
+//
+func (m PeerMethod1) Generate() (*DIDPeer, error) {
+	// Create a Genesis Version
+	return nil, util.NotImplementedError
+}
+
+func (d DIDPeer) GetMethodId() (string, error) {
```
decentralgabe, line 500: nit: `GetMethodID`

did/peer.go
```diff
@@ -0,0 +1,539 @@
...
+	m := string(d[9])
+	if _, ok := availablePeerMethods[m]; ok {
+		return m, nil
+	}
+	return "", errors.New(fmt.Sprintf("%s method not supported", m))
+}
+
+// id:peer:<method>
+func (d DIDPeer) GetMethod() (Resolver, error) {
```
decentralgabe, line 509: nit: `(*Resolver, error)` and `nil, error`s

decentralgabe (changes requested):
gave you a more thorough review this time. really close 😄

andorsk:
@decentralgabe thanks a ton for going through the PR again. Super appreciate the second pair of eyes. Definitely missed a few things which you caught. Any other issues lmk.

Again, thanks a ton for the review!

andorsk:
We should make a future issue where Resolver is actually re-specified to DIDResolver. There's a chance we have non DIDResolvers down the road and we'd want to not run into confusion/conflict here.

andorsk:
Think we are good, as I've addressed all the comments, but feel free to bring up more critiques. Comments are welcome 👍

PR #126, opened 2022-07-04
fix: remove build tags

andorsk:
related to https://github.com/lestrrat-go/jwx/issues/670 "You don't need to add build tags to your source code.".

PR #128, opened 2022-07-04
verifiable credential example

andorsk:
please merge https://github.com/TBD54566975/ssi-sdk/pull/126 first then this will follow: 

```
➜  vc git:(vc-example) go run vc.go
Created Verifiable Credential:
 {
  "@context": [
    "https://www.w3.org/2018/credentials/v1",
    "https://www.w3.org/2018/credentials/examples/v1"
  ],
  "id": "http://example.edu/credentials/1872",
  "type": [
    "VerifiableCredential",
    "AlumniCredential"
  ],
  "issuer": "https://example.edu/issuers/565049",
  "issuanceDate": "2010-01-01T19:23:24Z",
  "credentialSubject": {
    "alumniOf": {
      "id": "did:example:c276e12ec21ebfeb1f712ebc6f1",
      "name": [
        {
          "lang": "en",
          "value": "Example University"
        },
        {
          "lang": "fr",
          "value": "Exemple d'Université"
        }
      ]
    },
    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21"
  }```

### DIFF
--- a/examples/vc/vc.go
+++ b/examples/vc/vc.go
@@ -1,0 +1,74 @@
+// This is a very simple application which can create
+// a Verifiable Credential.
+// It can do so via a builder or directly
+// initializing a VerifiedCredentials struct
+// from the credentials package
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/TBD54566975/ssi-sdk/credential"
+	"github.com/TBD54566975/ssi-sdk/util"
+)
+
+func handleError(err error, msg string) {
+	if err != nil {
+		os.Stderr.WriteString(fmt.Sprintf("%s: %v", msg, err))
+		os.Exit(1)
+	}
+}
+
+func main() {
+
+	// Make a Verifiable Credential
+	// using the VC data type directly.
+	// Alternatively, use the builder
+	// A VC is set of tamper-evident claims and metadata
+	// that cryptographically prove who issued it
+	// Building a VC means using the CredentialBuilder
+	// as part of the credentials package in the ssk-sdk.
+	// VerifiableCredential is the verifiable credential model outlined in the
+	// vc-data-model spec https://www.w3.org/TR/2021/REC-vc-data-model-20211109/#basic-concept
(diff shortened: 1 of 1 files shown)


decentralgabe (commented):
@andorsk 
before moving your examples, let's agree on whether they should be tests or as is.

the main benefit of tests:
- not included in the binary when you're using the sdk as a dependency
- simple to run independently
- can easily determine if the examples break, as tests are run for a CI-precondition
- simpler error handling with assertions

benefits of main.go:
- easier to see "real world" usage

what do you think?

andorsk:
Me personally, I've suggested we follow precedence and standards, which is that these are not tests (b/c they aren't) and run as example files. There are more reasons I can give, but ultimately I don't think we are facing a unique problem here. Ultimately, how things have been done historically seems like a reasonable way to handle it. 

See here for some:
https://github.com/golang-standards/project-layout/tree/master/examples
https://github.com/cossacklabs/themis/tree/master/docs/examples/go

That being said: I'd suggest two lines here:

1. We can agree based upon the evidence above. But given we've both highlighted reasons and have not changed positions, it seems like it's at an impasse right now. 
2. We can ask some third person to weigh in and tie break this. 
3. If you feel strongly about this, I can change them to tests. As I mentioned before, It's more important we have something in for now. There are things I'm going to want to dig my heel in potentially later, but this isn't one of them. 

What's the most important here is we have some examples available for everyone else. We can always change things later if we realize we made a mistake here.

decentralgabe (approved):
@andorsk I appreciate the links, I wasn't aware this was a convention.

I'm OK to merge as-is. In the future, we can consider creating a new repo `ssi-sdk-examples` and moving them there; however for now this is an excellent start.

I noticed other examples rely on `handleError`, perhaps worth creating an examples util file or something similar.

andorsk:
Thanks @decentralgabe for merging this in. 

And yep! I think that's totally fair. In fact, maybe we should see about putting all the examples in the same directory. Rather than different ones.

Lots of work to do on the examples! I'm happy to chat about scope for improving them.

PR #133, opened 2022-07-05
update to latest cm schema

decentralgabe:
link relied upon for CM schema moved, breaking the tests


decentralgabe:
opened https://github.com/TBD54566975/ssi-sdk/issues/134 and https://github.com/TBD54566975/ssi-sdk/issues/135 as follow ups cc: @andorsk

PR #156, opened 2022-07-27
Fix remove print builder

andorsk:
removes the print statement and marshaling in the builder on the exchange layer. There is no good reason to force a print here.


andorsk:
going to resent a PR. fixing history.

PR #157, opened 2022-07-27
remove print and json marshalling in builder at exchange layer

andorsk:
removes the print statement and marshaling in the builder on the exchange layer. There is no good reason to force a print here i think.


decentralgabe (approved):
Oops! Thanks for getting this

PR #158, opened 2022-07-28
refactor crypto keys test

nitro-neal:
Removing duplicate code and dynamically running all supported key types


crypto/keys_test.go
```diff
@@ -1,209 +1,41 @@
...
+		})
+	}
+}
 
-	t.Run("RSA", func(tt *testing.T) {
-		pub, priv, err := GenerateKeyByKeyType(RSA)
-		assert.NoError(tt, err)
-		assert.NotEmpty(tt, pub)
-		assert.NotEmpty(tt, priv)
+func validateKeyGeneration(tt *testing.T, pub crypto.PublicKey, priv crypto.PrivateKey, keyType KeyType, err error) {
```
decentralgabe, line 19: I'm not sure I see the benefit in having this as a separate function.

Perhaps just embedding it in the above is sufficient?

crypto/keys_test.go
```diff
@@ -1,209 +1,41 @@
...
-		assert.NoError(tt, err)
-		assert.NotEmpty(tt, privKeyBytes)
-
-		reconstructedPriv, err := BytesToPrivKey(privKeyBytes, P521)
-		assert.NoError(tt, err)
-		assert.NotEmpty(tt, reconstructedPriv)
-		assert.EqualValues(tt, priv, reconstructedPriv)
-	})
+	for _, keyType := range GetSupportedKeyTypes() {
+		t.Run(string(keyType), func(tt *testing.T) {
```
decentralgabe, line 12: Table driven tests are a go convention.

What do you think about adjusting this from a loop to a table test format? 

https://github.com/golang/go/wiki/TableDrivenTests

crypto/keys_test.go
nitro-neal: Yup embedding is a good call

crypto/keys_test.go
nitro-neal: It seems like for a table driven test it has known input and output. With this unit test it doesn't really have a known output, so the table would only have on dimension an input.

I could explicitly create an array in the test to match this basically or we can keep using GetSupportedKeyTypes

https://github.com/TBD54566975/ssi-sdk/blob/main/crypto/models.go#L45

decentralgabe (approved):

PR #167, opened 2022-08-11
adding tests to example packages

nitro-neal:
This runs the examples on tests. If an example has an issue and crashes / exits the test will fail. 

This makes sure our examples don't break and go unnoticed.


decentralgabe (approved):
easy!

nitro-neal:
image

PR #172, opened 2022-08-23
DID Web tests use mock HTTP Resolvers

### LINKED ISSUE #171: Fix did:web test
Test run failed https://github.com/TBD54566975/ssi-sdk/runs/7975156768?check_suite_focus=true
due to reliance on Spruce's did:web file.

We should not have external resolution for these tests - we should use mocks.

decentralgabe:
Fixes #171


nitro-neal (approved):
Looks good

PR #174, opened 2022-08-25
Credential Manifest - Update to Credential Response

decentralgabe:
Draft that needs to have the latest PE spec before continuing


decentralgabe:
based on https://github.com/decentralized-identity/credential-manifest/issues/110

PR #178, opened 2022-08-31
Add gomobile

nrobi144:
### Description
It opens the door for generating Java and Obj-C bindings. Ultimately my goal is to continuously provide feedback about the usability of the **SSI-SDK** on mobile.

My next steps would be a greenfield app with examples that demonstrate all of the standards contained by the SSI-SDK (for ex. resolving identifiers, encrypting/decrypting, signing, etc...)

Here's an example using `gomobile bind` to convert the `crypto` package  to Java and Obj-C
Screenshot 2022-08-31 at 12 06 41

Screenshot 2022-08-31 at 12 30 56


decentralgabe:
Great start @nrobi144. Do you think it's worth adding a Mage script (e.g. `mage mobile`) to run the generation command?
And perhaps adding a doc or section to the README about using this functionality?

nrobi144:
> Great start @nrobi144. Do you think it's worth adding a Mage script (e.g. `mage mobile`) to run the generation command? And perhaps adding a doc or section to the README about using this functionality?

@decentralgabe 💯 will do that today

nrobi144:
@decentralgabe I've added the `ios`, `android` and `mobile` commands to Mage. This is my first exposure to **_Go_** and **_Mage_**, so any suggestions are welcome (especially on the error handling part).

FYI: I'm currently building the `crypto`, `cryptosuite` and `did` packages (those seemed good candidates for the API). So far only `crypto` seems to have *exported names*, is that because of the early stage of the SDK?

decentralgabe:
> @decentralgabe I've added the `ios`, `android` and `mobile` commands to Mage. This is my first exposure to **_Go_** and **_Mage_**, so any suggestions are welcome (especially on the error handling part).
> 
> FYI: I'm currently building the `crypto`, `cryptosuite` and `did` packages (those seemed good candidates for the API). So far only `crypto` seems to have _exported names_, is that because of the early stage of the SDK?

I'm open to making more of the methods exported! I tried to keep the public API minimal and useful; however if there is utility to be had by making more public methods let's do it.

decentralgabe (approved):
excellent work!

ALRubinger:
@nrobi144 Thank you so much! This is awesome!

nrobi144:
> 

That makes sense. I'd start with mirroring the [ssi-sdk example package](https://github.com/TBD54566975/ssi-sdk/tree/main/example) on mobile with a simplified UI. @decentralgabe what do you think? I assume examples will demo the API layer of the SSI-SDK so that could guide us what needs to be exported? 🤔

decentralgabe:
> > 
> 
> That makes sense. I'd start with mirroring the [ssi-sdk example package](https://github.com/TBD54566975/ssi-sdk/tree/main/example) on mobile with a simplified UI. @decentralgabe what do you think? I assume examples will demo the API layer of the SSI-SDK so that could guide us what needs to be exported? 🤔

yes I agree with that, and if you run into issues in your work, please open an issue here outlining which APIs you'll need to be made public and for which purpose and I can help get that done.

Issue #150, opened 2022-07-19
cli tool


andorsk:
build a cli tool to support various operations and. verifications. Inspiration can be seen here: https://www.youtube.com/watch?v=AVnCVzW0rkI

decentralgabe:
I like the idea. Guessing this should be a new project that uses the SDK?

andorsk:
@decentralgabe maybe? I'm a little worried about run-away repo management. There's the `cmd` folder, that we could use at the beginning and if it gets to be a large project move it over?

decentralgabe:
Yeah @andorsk agree, I think keeping it in a cmd directory here is a good first step.

andorsk:
started a PR here: https://github.com/TBD54566975/ssi-sdk/pull/183

decentralgabe:
Check out https://lists.w3.org/Archives/Public/public-credentials/2022Sep/0215.html

PR #183, opened 2022-09-06
added cli tool

andorsk:
added a cli tool for creating did keys.

```
~/workspace/github.com/andorsk/ssi-sdk/cmd $ go run cli.go create key
Chose  Ed25519
Generated DID Key:
did:key:z6MknCyPKLhv92CoHZsqJF1XHE6fchHKJfoqh26GAsCwUewD
```


decentralgabe:
Nice, great start.

I've never heard of `urface/cli` but I see it has a lot of usage! Most of my familiarity is with [cobra](https://github.com/spf13/cobra) (love/hate). I'd like us to consider the tradeoffs with these too and maybe some others before proceeding. Once we commit to an approach I want to be confident it'll carry us a long way.

I will try to write up my (brief) thoughts soon, but eager to hear what you think.

cmd/cli.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/urfave/cli/v2"
+)
+
+func main() {
+
+	app := &cli.App{
+		Commands: []*cli.Command{
+			{
+				Name:    "get",
+				Aliases: []string{"c"},
```
decentralgabe, line 20: what's this?

cmd/cli.go
```diff
@@ -0,0 +1,82 @@
...
+	"os"
+
+	"github.com/TBD54566975/ssi-sdk/crypto"
+	"github.com/TBD54566975/ssi-sdk/did"
+	"github.com/urfave/cli/v2"
+)
+
+func main() {
+
+	app := &cli.App{
```
decentralgabe, line 16: mostly thinking how does this scale to many commands? maybe a more modular structure

andorsk:
Honestly haha 🤷 !!! I don't know enough about the tradeoff's between urface/cli vs. cobra. I'd imagine there are some edge cases that one does better than the other, but for the general use case they both seem fine. 

https://zerokspot.com/weblog/2021/01/25/building-a-cli-using-urfave-cli/
https://mt165.co.uk/blog/golang-cli-library/

Seems like urface/cli  is more lightweight but less features? So I guess how bigs the scope of this cli tool? If it's simple, probably cli otherwise if this is going to be very large cobra?

cmd/cli.go
andorsk: yea....i was thinking we could refactor once we actually start to scale out features. Was a little concerned about causing unnecessary redirect when it's currently so simple. What do you think? modularize it now or later? Either way, at some point: agreed that needs to happen.

cmd/cli.go
andorsk: this is a get command. to get a list of supported types. Alias should probably be g not c though.

andorsk:
```
get typego run cli.go get key types
[Ed25519 X25519 secp256k1 P-224 P-256 P-384 P-521 RSA]
```

andorsk:
@decentralgabe thanks for the feedback! Added a few comments. Depending on where we land with some of these thoughts I'll address them accordingly and send a new review request.

nitro-neal:
I love the effort and this will be needed in the very near future. I think we should hold this pr for now and create a design doc that incorporates all the options necessary to spec out what this would look like.

I say this to make sure that it wont get messy later on.

Some initial thoughts on what should be included in the design doc:
* create dids with different methods (key,pkh,web,...)
* functionality to see if an input did is valid or not
* resolve vs expand functionality 
* support for special key functionality for different key methods

Eventually this CLI will be used for not just DIDS but everything from creating credential manifests, applications, vp and adding these dids to them.

Getting a clearer picture of how all components will tie together in the cli will be helpful.

andorsk:
@nitro-neal great idea. There is an issue open right now to discuss this https://github.com/TBD54566975/ssi-sdk/issues/150 but it hasn't gotten much action. I'm happy to iterate on a design doc with others, and then build out the initial PR to submit it to the repo. What do you think?

nitro-neal:
Yea! Maybe a good first pass would to update the readme (or make a doc.md in the same directory) to be more full featured kind of showing how it would look in finality with all did functionality and other entities and how a did would be used on those other entities.


We do a formal process for the SSI-Service for inspiration: https://github.com/TBD54566975/ssi-service/tree/main/sip/sips/sip_2 

Definitely doesn't have to be as formal or heavy but something that shows the complete picture.

andorsk:
Yea...good idea. I can take that.

andorsk:
Started outlining a document here: https://github.com/TBD54566975/ssi-sdk/pull/185 but have some work to do before it's ready to review.

andorsk:
putting this on draft until #185 is finished and accepted.

michaelneale:
hey @andorsk @decentralgabe would be super keen to have this (I was looking for something this week) - any chance we can move it along?

andorsk:
sorry for the late response, somehow i missed this: @michaelneale. 

unfortunately this is pretty low priority for me right now but I'll start peppering in some updates so at least it stops stagnating and starts moving forward. Thanks for letting me know this is something you'd care about.

michaelneale:
@andorsk oh no worries at all - no hurry on this, just want to make sure it is kept alive for when the time comes.

PR #184, opened 2022-09-07
Mobile wrapper test

decentralgabe:


mobile/crypto.go
```diff
@@ -0,0 +1,144 @@
...
+	supported := GetSupportedSignatureAlgs()
+	for _, a := range supported {
+		if sa == a {
+			return true
+		}
+	}
+	return false
+}
+
+func GetSupportedSignatureAlgs() []string {
```
nrobi144, line 33: @decentralgabe seems like arrays aren't yet supported https://github.com/golang/go/issues/13445

I took a stab at this with
```
package util

type StringArray interface {
	Add(s string) StringArray
	Get(i int) string
	Size() int
}

type StringArrayImpl struct {
	items []string
}

func (stringArray StringArrayImpl) Add(s string) StringArray{
	stringArray.items = append(stringArray.items, s)
	return stringArray
}

func (stringArray StringArrayImpl) Get(i int) string {
	return stringArray.items[i]
}

func (stringArray StringArrayImpl) Size() int {
	return len(stringArray.items)
}

func ToStringArray(s []string) StringArray {
	return StringArrayImpl{items: s}
}
```

where

```
func GetSupportedKeyTypes() util.StringArray {
	return util.ToStringArray([]string{keyTypeToString(ssi.Ed25519), keyTypeToString(ssi.X25519), keyTypeToString(ssi.Secp256k1), keyTypeToString(ssi.P224), keyTypeToString(ssi.P256), keyTypeToString(ssi.P384), keyTypeToString(ssi.P521), keyTypeToString(ssi.RSA)})
}
```

nrobi144: FYI: I've pushed this to https://github.com/nrobi144/ssi-sdk-mobile/blob/main/crypto/crypto.go

nrobi144: Still investigating why the `GenerateXYKey()` functions are failing, I can see both `ECDSAKeyPair` and `CryptoKeyPair`

Screenshot 2022-09-07 at 13 28 55

nrobi144: Something's up with `struct type`s. 

Changing `func GenerateX25519Key() (CryptoKeyPair, error)` to `func GenerateX25519Key() (string, error)` works.

Even though `gomobile` has these rules:
> - Any function type all of whose parameters and results have supported types. Functions must return either no results, one result, or two results where the type of the second is the built-in 'error' type.
> - Any struct type, all of whose exported methods have supported function types and all of whose exported fields have supported types.

nrobi144: Update: using pointers for `struct`s works:
For ex:
```
func GenerateEd25519Key() (*CryptoKeyPair, error) {
	privKey, pubKey, err := ssi.GenerateEd25519Key()
	return &CryptoKeyPair{
		PrivKey: privKey,
		PubKey:  pubKey,
	}, err
}
```

decentralgabe: thanks @nrobi144 I'm going to close this PR in favor of your projects.

let me know if I can help with more go-specific stuff

PR #192, opened 2022-09-14
DID pkg test coverage

decentralgabe:
Increasing test coverage. Minor refactors.


decentralgabe:
actually found a few issues!

did/key.go
```diff
@@ -31,6 +31,23 @@ const (
 	DIDKeyPrefix = "did:key"
 )
 
+func (d DIDKey) IsValid() bool {
+	return true
```
nitro-neal, line 35: no checks here?

nitro-neal (commented):
looks great

did/key.go
nitro-neal, line 35: no checks for this?

nitro-neal (commented):
look great!

nitro-neal (approved):

did/key.go
decentralgabe: updated, whoops

PR #193, opened 2022-09-15
trying out govuln

michaelneale:
Go has build in vuln management - thought it may be worth a try

https://go.dev/blog/vuln


michaelneale:
@decentralgabe not familiar enough with permissions to know how to try tweaks from a fork to workflows (I would hope it doesn't allow them)

decentralgabe (changes requested):
thanks @michaelneale this is great to add. I think it should be done via mage, perhaps a `mage vuln` check?

assuming that if it finds a vuln it'll return an error code and fail CI

michaelneale:
yes I think could be part of mage (still not sure if govuln should be run earlier or later, need to read more about how it works). If its always quick, makes sense to have it earlier before tests (or in parallel perhaps?)

decentralgabe:
> yes I think could be part of mage (still not sure if govuln should be run earlier or later, need to read more about how it works). If its always quick, makes sense to have it earlier before tests (or in parallel perhaps?)

parallel would be nice, but the pipeline isn't too long now a days, so I don't mind it going first, last, or wherever!

decentralgabe:
Realized this relies on go 1.18. We're still on 1.17, which is pre-generics support. Given that 1.19 is out it seems reasonable to bump to the latest. Though I'm still a little hesitant to start introducing generics 🤔 

Opened https://github.com/TBD54566975/ssi-sdk/issues/198

decentralgabe:
ok, #198 is done - after mage this should be good to go

michaelneale:
@decentralgabe ah thanks. Hadn't had a chance to look at this again. This tool seems nice - it won't return an error if there are informational only warnings (which there are at the moment - it knows that code isn't used). I think it could run parallel to build step perhaps? or do you like it how it is.

decentralgabe:
@michaelneale let's go for parallel!

michaelneale:
@decentralgabe ok - try it again (in parallel - doesn't need to run the build to scan).

decentralgabe:
nice @michaelneale looks like it works

decentralgabe (approved):
approval pending mageification of govuln

michaelneale:
@decentralgabe not sure what you mean by mageification?

michaelneale:
@decentralgabe ah thanks. Trying it again with mage - letting it install govulncheck on demand is probably a good way to go to keep it fresh too (and easier for end users).

decentralgabe:
Screen Shot 2022-09-27 at 8 30 35 AM

Hmm looks like the binary isn't found on the path 🤔 : https://github.com/TBD54566975/ssi-sdk/actions/runs/3120487785/jobs/5094376835

decentralgabe:
@michaelneale I pushed a fix to your branch https://github.com/michaelneale/ssi-sdk/pull/3

threw in some other magefile tidyings, couldn't help myself 😉

decentralgabe:
the issue is the command needs `go install` but the magefile was using `go get`

Issue #194, opened 2022-09-15
JWT implementation is tightly coupled to the Linked Data Suite Implementation


decentralgabe:
Today, the SDK requires you to use a `JSONWebKeySigner` which is tightly coupled to the `JsonWebSignature` suite.

This ask decouples the two, allowing for a 'pure' JWT implementation

PR #195, opened 2022-09-15
Decouple JWT and LD signing/verifying paths

decentralgabe:
fixes https://github.com/TBD54566975/ssi-sdk/issues/194

Notably, this makes all endpoints use JWTs, leaving LD suites to be a bit lonely. LD suite support is *possible* but not integrated. Still thinking of the best way to include this...


decentralgabe:
Added https://github.com/TBD54566975/ssi-sdk/issues/196 to address

crypto/jwt.go
```diff
@@ -0,0 +1,182 @@
...
+func (sv *JWTSigner) GetSigningAlgorithm() string {
+	return sv.Algorithm()
+}
+
+// SignJWT takes a set of JWT keys and values to add to a JWT before singing them with the key defined in the signer
+func (sv *JWTSigner) SignJWT(kvs map[string]interface{}) ([]byte, error) {
+	t := jwt.New()
+	for k, v := range kvs {
+		if err := t.Set(k, v); err != nil {
+			err := errors.Wrapf(err, "could not set %sv to value: %v", k, v)
```
nitro-neal, line 89: %sv did you mean %v ?

decentralgabe, line 89:
```suggestion
			err := errors.Wrapf(err, "could not set %s to value: %v", k, v)
```

decentralgabe: %s, since the key is a string. fixed.

PR #215, opened 2022-10-05
Writing a new end to end steelthread use case

nitro-neal:
Left 

image

```
// Annotated steel thread flow for calling out all signing, verification, and key management.
// This example is focused on the signing, exchange, and verification of objects.

// We assume there are two parties:
//	1 . Alice, using a wallet, applying for a credential
//	2. An issuer via the SSI Service, processing credential applications and issuing credentials

// Each party has a single DID. Alice and the SSI Service each have a single did:key DID.
// Alice stores her DID’s private key in her wallet.
// The SSI Service stores its private key in the service key store database.
```
Note:
Leaving a few 'TODOs' inline that will be implemented when we have signing and the real credential objects to submit


example/usecase/steel_thread/steel_thread.go
```diff
@@ -0,0 +1,320 @@
...
+type Entity struct {
+	didKey                did.DIDKey
+	verifier              crypto.JWTVerifier
+	credentialManifest    manifest.CredentialManifest
+	credentialApplication manifest.CredentialApplication
+	credentialResponse    manifest.CredentialResponse
+	verifiableCredentials []credential.VerifiableCredential
+}
+
+func (t *Entity) GenerateWallet() {
```
decentralgabe, line 35: there is a wallet in `/example/wallet.go` - is that usable here?

decentralgabe (approved):
excellent!

Issue #213, opened 2022-10-04
[Idea] Customizable Verifiers


decentralgabe:
Many of the objects in the SDK need to be verified. This process can include steps such as:

1. Making sure a given signature is valid
2. Making sure the data complies with the spec object model
3. Making sure the data is not expired or revoked
4. Making sure the data is issued from/to expected parties
5. Making sure the data complies with a given schema
6. Probably more!

The idea is to create a Verifier, possibly using a builder pattern, that allows a developer to build "verifiers" for each object they wish to verify, including whichever checks they care about - and having the option of adding additional checks not provided by this project.

decentralgabe:
Done in #217

PR #217, opened 2022-10-06
Dynamic Credential Verification

decentralgabe:
#213 

Left out signing, covered everything else.


credential/verification/verification.go
```diff
@@ -0,0 +1,72 @@
...
+	OptionKey string
+)
+
+// VerificationOption represents a single option that may be required for a verifier
+type VerificationOption struct {
+	ID     OptionKey
+	Option interface{}
+}
+
+// GetVerificationOption	returns a verification option given an ID
```
nitro-neal, line 31: Spacing issue in comment

credential/verification/verification.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	return nil, errors.Errorf("option with id <%s> not found", id)
+}
+
+type Verify func(credential credential.VerifiableCredential, opts ...VerificationOption) error
+
+// NewCredentialVerifier creates a new credential verifier which executes in the order of the verifiers provided
+func NewCredentialVerifier(verifiers []Verifier) (*CredentialVerifier, error) {
+	// dedupe
+	var res []Verifier
```
nitro-neal, line 46: did you mean this variable name?

credential/verification/verifiers.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/pkg/errors"
+)
+
+const (
+	SchemaOption OptionKey = "schema"
+)
+
+var KnownVerifiers = []Verifier{
+	{
+		ID:         "Go Validation",
```
nitro-neal, line 18: instead of Go Validation maybe like Object Validation?

credential/verification/verifiers.go
```diff
@@ -0,0 +1,94 @@
...
+// VerifyJSONSchema verifies a credential's data against a Verifiable Credential JSON Schema:
+// https://w3c-ccg.github.io/vc-json-schemas/v2/index.html#credential_schema_definition
+// There is a required single option which is a string JSON value representing the Credential Schema Object
+func VerifyJSONSchema(credential credential.VerifiableCredential, opts ...VerificationOption) error {
+	hasSchemaProperty := credential.CredentialSchema != nil
+	schema, err := GetVerificationOption(opts, SchemaOption)
+	if err != nil {
+		// if the credential does not have a schema property, we cannot perform this check
+		if !hasSchemaProperty {
+			return nil
```
nitro-neal, line 70: Is this expected functionality?

If we you VerifyJSONSchema and it has no schema it will return nil (no error)

nitro-neal (approved):
Would there ever be a case where we would not run all verifiers? 

Would a system admin like omit some verifiers on certain creds?

credential/verification/verification.go
decentralgabe: updated

credential/verification/verifiers.go
decentralgabe: yes, updated

credential/verification/verifiers.go
decentralgabe: there is a test for this case - if you do not pass in a schema AND the credential does not have a schema this will pass

here are the cases:
- no schema property, no schema provided = no error
- no schema property, schema provided = error
- schema property, no schema provided = error
- schema property, schema provided = depends on check

I think this is expected.

decentralgabe:
> Would there ever be a case where we would not run all verifiers?

I think so. The go object model may not encapsulate all the fields that a verifier wishes to check, so they may deviate here. There also may be use cases where a verifier wants a credential even if it is expired - to prove you once had a credential, perhaps for a re-issuance flow.

> Would a system admin like omit some verifiers on certain creds?

I think so, yes. I imagine there can be verifiers defined per credential grouping. How that grouping is defined...not exactly sure. Maybe by schema, or by issuer. This could be functionality we add to the service.

decentralgabe:
merging, if you have follow ups I can take them in another PR @nitro-neal

PR #218, opened 2022-10-07
move did resolver to a new file, make method public

decentralgabe:


decentralgabe:
bump minor version of go

PR #220, opened 2022-10-07
LD Key utilities

decentralgabe:


crypto/models.go
```diff
@@ -32,6 +32,10 @@ const (
 	PS256 SignatureAlgorithm = "PS256"
 )
 
+func (kt KeyType) String() string {
```
nitro-neal, line 35: I'm more used to .ToString() (java brain here) but if this is the better Go convention we can stick with this

cryptosuite/jsonwebkey2020.go
```diff
@@ -270,3 +275,23 @@ func NewJSONWebKeyVerifier(kid string, key crypto.PublicKeyJWK) (*JSONWebKeyVeri
...
+	case crypto.Ed25519.String(), did.Ed25519VerificationKey2018.String(), did.Ed25519VerificationKey2020.String():
+		convertedKeyType = crypto.Ed25519
+	case crypto.X25519.String(), did.X25519KeyAgreementKey2019.String(), did.X25519KeyAgreementKey2020.String():
+		convertedKeyType = crypto.X25519
+	case crypto.SECP256k1.String(), did.EcdsaSecp256k1VerificationKey2019.String():
+		convertedKeyType = crypto.SECP256k1
+	default:
+		return nil, fmt.Errorf("unsupported key type: %s", kt)
+	}
+	return crypto.BytesToPubKey(keyBytes, convertedKeyType)
```
nitro-neal, line 295: type {
  keyBytes []byte
  keyType kt.String() type
}

would a wrapper object give much value? I guess you already have the keytype in hand so maybe not

nitro-neal (approved):

crypto/models.go
decentralgabe: yeah, I think it is more of a go thing:
- https://github.com/xeipuuv/gojsonschema/blob/master/result.go#L68
- https://github.com/lestrrat-go/jwx/blob/ec54ec431aa503ab1d5fc11a5c97a70daf2e0171/jwt/options_gen_test.go#L12

cryptosuite/jsonwebkey2020.go
decentralgabe: hmm...yeah we have `LDKeyType` and this is going to `KeyType`, but the interface that this method is most often being used with is expecting a `gocrypto.PublicKey`

Issue #134, opened 2022-07-05
Have the ability to cache / load local copies of referenced schemas


decentralgabe:
Some schemas, such as those for [Credential Manifest](https://github.com/TBD54566975/ssi-sdk/blob/main/schema/known_schemas/cm-credential-manifest.json#L59) rely on additional schemas using the `$ref` property.

Utilizing these schemas in the sdk resolves in network calls, to resolve the `$ref` values. It's possible some of these schema URIs may become unavailable. We should have a method to "locally" resolve such schema references. This is possible in [ajv](https://ajv.js.org/api.html#ajv-addschema-schema-object-object-key-string-ajv).

nitro-neal:
First part of this - https://github.com/TBD54566975/ssi-sdk/pull/221

nitro-neal:
Added:

// DerefSchemas takes our known schemas and dereferences the schema's $ref http links to be a part of the json schema object.
// This makes our code faster when doing validation checks and allows us to not ping outside sources for schemas refs which may go down or change.
// TODO: (Neal) Currently we do not use these dereferenced schemas in code because there is more work to be done here.
// Currently these dereferenced schemas are missing some information and fail validation with our known json objects
// I believe some more work in the investigation library needs to be done and we need to handle circular dependencies

it could also be our logic to do validation is strange, it seems that these schemas + json objects with an online tool check out..

![image](https://user-images.githubusercontent.com/5314059/194940695-01845912-825f-406c-a6b5-91ce1c1404c9.png)

PR #221, opened 2022-10-10
adding new mage target to dereference schemas

nitro-neal:
image

By running mage derefschemas this will create a new file (ex: cm-credential-manifest-deref.json) which will be the known schemas but fully dereferenced.

connected to this issue - https://github.com/TBD54566975/ssi-sdk/issues/134


decentralgabe:
@nitro-neal nice work. how can the dereferenced schemas be used in the code?

perhaps a short doc above the method describing the need and purpose? or in the README somewhere?

nitro-neal:
Added:

// DerefSchemas takes our known schemas and dereferences the schema's $ref http links to be a part of the json schema object.
// This makes our code faster when doing validation checks and allows us to not ping outside sources for schemas refs which may go down or change.
// TODO: (Neal) Currently we do not use these dereferenced schemas in code because there is more work to be done here.
// Currently these dereferenced schemas are missing some information and fail validation with our known json objects
// I believe some more work in the investigation library needs to be done and we need to handle circular dependencies

it could also be our logic to do validation is strange, it seems that these schemas + json objects with an online tool check out..
image

decentralgabe (approved):
@nitro-neal given that the schemas need some more work would you prefer this PR to stay open, or will you be taking the additional changes in a follow up PR?

Your choice. Approving.

Issue #62, opened 2022-03-21
Support Decentralized Web Node Messaging


decentralgabe:
https://identity.foundation/decentralized-web-node/spec/#messages

decentralgabe:
Partially handled in #210

decentralgabe:
Need to add schema validation for schemas here - https://github.com/TBD54566975/dwn-sdk-js/tree/main/src/validation/json-schemas

nitro-neal:
We dwn messages for now because this use case is now handled in the relay service

PR #224, opened 2022-10-11
schema and pkh refactor

decentralgabe:
- update credential manifest schema from https://github.com/decentralized-identity/credential-manifest/pull/122
- minor pkh refactor
- moved out non-schemas from schema dir


did/pkh.go
```diff
@@ -29,24 +33,29 @@ const (
...
+	EthereumNetworkPrefix = "eip155:1"
+	PolygonNetworkPrefix  = "eip155:137"
 
-// The following context should be manually inserted into each DID Document. This will likely change
-// over time as new verification methods are supported, and general-purpose methods are specified.
-var knownDIDPKHContext, _ = schema.GetKnownSchema(pkhContextFilename)
+	EcdsaSecp256k1RecoveryMethod2020 = "EcdsaSecp256k1RecoveryMethod2020"
+)
+
+// GetDIDPKHContext returns a context  which should be manually inserted into each did:pkh document. This will likely
```
nitro-neal, line 44: two spaces between context and which

did/pkh.go
```diff
@@ -82,14 +91,58 @@ func (d DIDPKH) Method() Method {
...
+		if strings.Contains(did, prefix+":") {
+			return GetDIDPKHNetworkForPrefix(prefix)
 		}
 	}
-	return nil, util.LoggingNewError("network not supported")
+	return "", fmt.Errorf("could not find network for did:pkh DID: %s", did)
+}
+
+// GetVerificationTypeForNetwork returns the verification key type for a given network
+func GetVerificationTypeForNetwork(n Network) (string, error) {
```
nitro-neal, line 132: Nice, we don't have to have a mapping anymore, there could be a case where we add a new pkh but forget to add it here in the GetVerificationTypeForNetwork function, thats why I opt to have one source of truth mapping. But this works good too

nitro-neal (approved):
Looks good!

dwn/model.go
```diff
@@ -53,27 +47,28 @@ func (msg *DWNMessage) IsValid() error {
...
-		return errors.Wrap(err, "could not get dwn message schema")
-	}
-	if err = schema.IsJSONValidAgainstSchema(string(jsonBytes), s); err != nil {
-		logrus.WithError(err).Errorf("dwn message not valid against schema")
-		return err
-	}
-	return nil
+// IsValidDWNMessage validates a given dwn message object against its known JSON schema
+func IsValidDWNMessage(msg DWNMessage) error {
+	// TODO: add a schema when there is one
```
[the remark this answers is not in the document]
decentralgabe: changing this to link to https://github.com/TBD54566975/ssi-sdk/issues/62 to re-add schemas

did/pkh.go
decentralgabe: good point, let me see how to define this once

PR #230, opened 2022-10-17
adding vc validation to IsValidCredentialApplicationForManifest function

nitro-neal:
adding vc validation to IsValidCredentialApplicationForManifest function


credential/manifest/model.go
```diff
@@ -144,8 +145,9 @@ func (cf *CredentialResponse) IsValid() error {
 }
 
 // IsValidCredentialApplicationForManifest validates the rules on how a credential manifest [cm] and credential application [ca] relate to each other https://identity.foundation/credential-manifest/#credential-application
-func IsValidCredentialApplicationForManifest(cm CredentialManifest, ca CredentialApplication) error {
+func IsValidCredentialApplicationForManifest(cm CredentialManifest, ca CredentialApplication, vcs []credential.VerifiableCredential) error {
```
decentralgabe, line 148: nit: consider making this a variadic argument so it can be an optional parameter 

```
IsValidCredentialApplicationForManifest(cm CredentialManifest, ca CredentialApplication, vcs ...credential.VerifiableCredential)
```

credential/manifest/model.go
```diff
@@ -212,11 +224,80 @@ func IsValidCredentialApplicationForManifest(cm CredentialManifest, ca Credentia
...
+
+		// index submission descriptors by id of the input descriptor
+		submissionDescriptorLookup := make(map[string]exchange.SubmissionDescriptor)
+		for _, d := range ca.PresentationSubmission.DescriptorMap {
+			submissionDescriptorLookup[d.ID] = d
+		}
+
+		// validate each input descriptor is fulfilled
+		for _, inputDescriptor := range cm.PresentationDefinition.InputDescriptors {
+			submissionDescriptor, ok := submissionDescriptorLookup[inputDescriptor.ID]
```
decentralgabe, line 237: at this point in time how will the fulfilled input descriptors have been marked as valid?

credential/manifest/model.go
```diff
@@ -212,11 +224,80 @@ func IsValidCredentialApplicationForManifest(cm CredentialManifest, ca Credentia
...
+
+			// resolve the claim from the JSON path expression in the submission descriptor
+			submittedClaim, err := jsonpath.JsonPathLookup(vcs, submissionDescriptor.Path)
+
+			if err != nil {
+				return errors.Wrapf(err, "could not resolve claim from submission descriptor<%s> with path: %s", submissionDescriptor.ID, submissionDescriptor.Path)
+			}
+
+			// convert submitted claim vc to map[string]interface{}
+			cred := submittedClaim.(credential.VerifiableCredential)
```
decentralgabe, line 263: does this work reliably? I always did the json marshal/unmarshal if this works it could simplify a good amount of code

credential/manifest/model.go
```diff
@@ -212,11 +224,80 @@ func IsValidCredentialApplicationForManifest(cm CredentialManifest, ca Credentia
...
+
+			if err := cred.IsValid(); err != nil {
+				return errors.Wrap(err, "vc is not valid")
+			}
+
+			var vcMap map[string]interface{}
+			credJson, _ := json.Marshal(cred)
```
decentralgabe, line 270: `credJSOL` and let's check this error, and the one on the line below

credential/manifest/model_test.go
```diff
@@ -215,61 +221,62 @@ func TestIsValidPair(t *testing.T) {
...
 
 		ca.PresentationSubmission.DescriptorMap[0].Path = "bad-path"
 
-		err = IsValidCredentialApplicationForManifest(cm, ca)
+		err = IsValidCredentialApplicationForManifest(cm, ca, vcs)
 		assert.Contains(t, err.Error(), "invalid json path: bad-path")
 
 	})
+
+	t.Run("Credential Application and Credential Manifest Pair VC id test", func(tt *testing.T) {
```
decentralgabe, line 261: it might be good to add a few specific types of input descriptors to check for - e.g. a test with a mismatched format, a test with a mismatched path, one where only some of the input descriptors were fulfilled, cases where multiple creds can fulfill multiple input descriptors, another where only one cred can fulfill multiple input descriptors...

credential/manifest/model.go
nitro-neal: Hmm, so if there is no vcs then would it be expected that the application's presentation_submission's descriptor_map be empty? 
image

credential/manifest/model.go
nitro-neal: well I think at this point we don't even know they are fulfilled (that the last part of the for loop)

This for loop
checks if the IDs match up
checks if the formats match up
makes sure we can resolve json path

credential/manifest/model.go
nitro-neal: seems to work in all unit tests, I will create more tests

credential/manifest/model_test.go
nitro-neal: Added more tests

credential/manifest/model.go
decentralgabe: ahhh misunderstood this - makes sense now

credential/manifest/model.go
decentralgabe: yes, exactly

decentralgabe: if there are input descriptors there must be VCs, so if there's a presentation definition at all - there must be VCs

nitro-neal: ok good stuff, I was trying to make a presentation definition with 0 input descriptors, 
but yup this works now with just an empty presentation definition

decentralgabe (approved):

PR #236, opened 2022-10-18
manifest / application checking by json path

decentralgabe:


credential/manifest/model.go
```diff
@@ -145,8 +155,9 @@ func (cf *CredentialResponse) IsValid() error {
 	return util.NewValidator().Struct(cf)
 }
 
-// IsValidCredentialApplicationForManifest validates the rules on how a credential manifest [cm] and credential application [ca] relate to each other https://identity.foundation/credential-manifest/#credential-application
-func IsValidCredentialApplicationForManifest(cm CredentialManifest, ca CredentialApplication, vcs ...credential.VerifiableCredential) error {
+// IsValidCredentialApplicationForManifest validates the rules on how a credential manifest [cm] and credential
+// application [ca] relate to each other https://identity.foundation/credential-manifest/#credential-application
+func IsValidCredentialApplicationForManifest(cm CredentialManifest, ca CredentialApplication, applicationJSON map[string]interface{}) error {
```
nitro-neal, line 160: you sure you want to do it like this?

like what if it took in []VerifiableCredentials and then converted them to map[string]interface{} internally

also, we could get the ca CredentialApplication object out of the json too, so if we do accept the json string we could build a ca object and cs map[string]interface{} as the first ste[

nitro-neal: yea just remove this ca CredentialApplication and create this on the first line with the applicationJSON and then good to go

nitro-neal (approved):

credential/manifest/model.go
decentralgabe: done done

PR #238, opened 2022-10-19
For Credential Application processing, handle JWT and Data Integrity Creds

decentralgabe:


credential/manifest/model.go
```diff
@@ -322,6 +323,36 @@ func IsValidCredentialApplicationForManifest(cm CredentialManifest, applicationA
 	return nil
 }
 
+// turn a generic cred into a known shape without maintaining the proof/signature wrapper
+func credentialsFromInterface(genericCred interface{}) (*credential.VerifiableCredential, error) {
```
nitro-neal, line 327: awesome, like if we want to refactor one day, this function can be used at the very first line in the service.

The service input gives us 'something' we instantly take that something and turn it into a *credential.VerifiableCredential. We can do this for all objects that can take different forms

nitro-neal: so maybe make it capital CredentialsFromInterface because I have a feeling we will use this somewhere else

decentralgabe: agreed...there is more refactoring to do here and moving a number of things from the service here

nitro-neal (approved):

credential/manifest/model.go
decentralgabe: moving to a new pkg in credentials called `util` to avoid import cycles

decentralgabe:
added tests

PR #243, opened 2022-10-26
fix postal address

nitro-neal:


decentralgabe (approved):

credential/exchange/builder_test.go
```diff
@@ -94,7 +95,60 @@ func TestPresentationDefinitionBuilder(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	})
+
+	assert.NoError(t, err)
+	definition, err := builder.Build()
+	assert.NoError(t, err)
+	println(util.PrettyJSON(definition))
```
andresuribe87, line 147: Is this println intentional?

decentralgabe, line 147:
```suggestion
```

decentralgabe: good catch - removed for Neal

credential/exchange/builder_test.go
```diff
@@ -94,7 +95,59 @@ func TestPresentationDefinitionBuilder(t *testing.T) {
...
+						},
+					},
+				},
+			},
+		},
+	})
+
+	assert.NoError(t, err)
+	definition, err := builder.Build()
+	assert.NoError(t, err)
```
decentralgabe, line 146:
```suggestion
	assert.NoError(t, err)
	assert.NotEmpty(t, definition)
```

PR #244, opened 2022-10-26
Support returning failed input descriptors on cred application validation

decentralgabe:


credential/manifest/model.go
```diff
@@ -209,130 +212,151 @@ func IsValidCredentialApplicationForManifest(cm CredentialManifest, applicationA
...
-	// The Credential Application object MUST contain a presentation_submission property IF the related Credential Manifest contains a presentation_definition.
-	// Its value MUST be a valid Presentation Submission:
-	if !cm.PresentationDefinition.IsEmpty() {
-		if ca.PresentationSubmission.IsEmpty() {
-			return errresp.NewErrorResponse(errresp.ApplicationError, "credential application's presentation submission cannot be empty because the credential manifest's presentation definition is not empty")
-		}
+	// The Credential Application object MUST contain a presentation_submission property IF the related Credential
+	// Manifest contains a presentation_definition. Its value MUST be a valid Presentation Submission:
+	if cm.PresentationDefinition.IsEmpty() {
+		return
```
nitro-neal, line 231: good call, 

I wonder if there is an edge case where if there is not a PresDef but the presentation submission exists (for some reason?) and then it is invalid

error/response.go
```diff
@@ -11,7 +11,7 @@ type (
 )
 
 const (
-	ApplicationError Type = "ApplicationError"
+	ApplicationError Type = "Valid"
```
nitro-neal, line 14: do you want to rename ApplicationError to ValidError?

nitro-neal (approved):

credential/manifest/model.go
decentralgabe: great point - will add a check

error/response.go
decentralgabe: whoops, will revert

PR #246, opened 2022-10-27
Enable golangci-lint

andresuribe87:
This PR includes:
* Adding mage `lint` and `deps` targets.
  * `lint` will run golangci-lint with the commited configuration. I disabled a number of checks for revive, and updated the source code so that they all passed. 
  * `deps` will install `golangci-lint` using homebrew.
* Added the `golangci-lint` action according to the [instructions](https://github.com/golangci/golangci-lint-action#how-to-use)

### DIFF
--- a/credential/manifest/validation.go
+++ b/credential/manifest/validation.go
@@ -16,41 +16,43 @@ import (
 // IsValidCredentialApplicationForManifest validates the rules on how a credential manifest [cm] and credential
 // application [ca] relate to each other https://identity.foundation/credential-manifest/#credential-application
 // applicationAndCredsJSON is the credential application and credentials as a JSON object
-func IsValidCredentialApplicationForManifest(cm CredentialManifest, applicationAndCredsJSON map[string]interface{}) (unfulfilledInputDescriptors map[string]string, err error) {
+func IsValidCredentialApplicationForManifest(cm CredentialManifest, applicationAndCredsJSON map[string]interface{}) (map[string]string, error) {
+	var err error
+
 	// parse out the application to its known object model
 	applicationJSON, ok := applicationAndCredsJSON[CredentialApplicationJSONProperty]
 	if !ok {
 		err = errresp.NewErrorResponse(errresp.ApplicationError, "credential_application property not found")
-		return
+		return nil, err
 	}
 
 	applicationBytes, err := json.Marshal(applicationJSON)
 	if err != nil {
 		err = errresp.NewErrorResponseWithErrorAndMsg(errresp.CriticalError, err, "failed to marshal credential application")
-		return
+		return nil, err
 	}
 	var ca CredentialApplication
 	if err = json.Unmarshal(applicationBytes, &ca); err != nil {
 		err = errresp.NewErrorResponseWithErrorAndMsg(errresp.CriticalError, err, "failed to unmarshal credential application")
-		return
+		return nil, err
 	}
 
 	// Basic Validation Checks
 	if err = cm.IsValid(); err != nil {
 		err = errresp.NewErrorResponseWithErrorAndMsg(errresp.ApplicationError, err, "credential manifest is not valid")
-		return
+		return nil, err
 	}
 
 	if err = ca.IsValid(); err != nil {
 		err = errresp.NewErrorResponseWithErrorAndMsg(errresp.ApplicationError, err, "credential application is not valid")
-		return
+		return nil, err
 	}
 
 	// The object MUST contain a manifest_id property. The value of this property MUST be the id of a valid Credential Manifest.
 	if cm.ID != ca.ManifestID {
 		err = errresp.NewErrorResponsef(errresp.ApplicationError, "the credential application's manifest id: "+
 			"%s must be equal to the credential manifest's id: %s", ca.ManifestID, cm.ID)
-		return
+		return nil, err
 	}
 
 	// The ca must have a format property if the related Credential Manifest specifies a format property.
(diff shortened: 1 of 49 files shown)


credential/manifest/validation.go
```diff
@@ -16,41 +16,43 @@ import (
 // IsValidCredentialApplicationForManifest validates the rules on how a credential manifest [cm] and credential
 // application [ca] relate to each other https://identity.foundation/credential-manifest/#credential-application
 // applicationAndCredsJSON is the credential application and credentials as a JSON object
-func IsValidCredentialApplicationForManifest(cm CredentialManifest, applicationAndCredsJSON map[string]interface{}) (unfulfilledInputDescriptors map[string]string, err error) {
+func IsValidCredentialApplicationForManifest(cm CredentialManifest, applicationAndCredsJSON map[string]interface{}) (map[string]string, error) {
```
decentralgabe, line 19: I used named return here because I thought that a map[string]string on its own might not be clear. I agree this is a cleaner impl though

credential/verification/verifiers.go
```diff
@@ -30,19 +30,19 @@ var KnownVerifiers = []Verifier{
...
-	return credential.IsValid()
+func VerifyValidCredential(cred credential.VerifiableCredential, _ ...VerificationOption) error {
+	return cred.IsValid()
 }
 
 // VerifyExpiry verifies a credential's expiry date is not in the past. We assume the date is parseable as
 // an RFC3339 date time value.
-func VerifyExpiry(credential credential.VerifiableCredential, _ ...VerificationOption) error {
-	if credential.ExpirationDate == "" {
+func VerifyExpiry(verifiableCred credential.VerifiableCredential, _ ...VerificationOption) error {
```
decentralgabe, line 39: cred?

crypto/jwk_test.go
```diff
@@ -14,12 +14,12 @@ func TestJWKToPrivateKeyJWK(t *testing.T) {
 	assert.NotEmpty(t, privateKey)
 
 	// convert to JWK
-	jwk, err := jwk.New(privateKey)
+	jwkKey, err := jwk.New(privateKey)
```
decentralgabe, line 17: jwk means json web key, so this may be repetitive

decentralgabe: key alone is fine

cryptosuite/jsonwebkey2020.go
```diff
@@ -21,7 +21,7 @@ type (
 )
 
 const (
-	JsonWebKey2020 LDKeyType = "JsonWebKey2020"
+	JSONWebKey2020Name LDKeyType = "JsonWebKey2020"
```
decentralgabe, line 24: instead of name, maybe `JSONWebKey2020Type` ?

example/usecase/employer_university_flow/pkg/util.go
```diff
@@ -39,35 +39,34 @@ func GenerateDIDPeer() (did.DID, error) {
 	if err != nil {
 		return nil, err
 	}
-	did, err := did.PeerMethod0{}.Generate(kt, pubKey)
+	dID, err := did.PeerMethod0{}.Generate(kt, pubKey)
```
decentralgabe, line 42: think this should be reverted, it's either `did` or `DID`

example/usecase/employer_university_flow/pkg/util.go
```diff
@@ -39,35 +39,34 @@ func GenerateDIDPeer() (did.DID, error) {
...
 
 	var vc2 credential.VerifiableCredential
 	if err := util.Copy(&vc, &vc2); err != nil {
 		return err
 	}
 
-	var OKP = cryptosuite.KTY("OKP")
-	var Ed25519 = cryptosuite.CRV("Ed25519")
+	var okp = cryptosuite.KTY("OKP")
+	var ed25519 = cryptosuite.CRV("Ed25519")
```
decentralgabe, line 62:
```suggestion
	okp := cryptosuite.KTY("OKP")
	ed25519 := cryptosuite.CRV("Ed25519")
```

decentralgabe: actually, I think we already have these types defined. `cryptosuite.OKP` and `cryptosuite.Ed25519` should work

example/usecase/steel_thread/steel_thread.go
```diff
@@ -265,7 +261,7 @@ func createCredentialApplication(cm manifest.CredentialManifest) manifest.Creden
 	return credApp
 }
 
-func createVerifiableCredential(issuerDID string, walletDID string, descriptor manifest.OutputDescriptor) credential.VerifiableCredential {
+func createVerifiableCredential(issuerDID string, walletDID string, _ manifest.OutputDescriptor) credential.VerifiableCredential {
```
decentralgabe, line 264: if this isn't used, let's remove it

decentralgabe (changes requested):
few small ones - thanks for adding this!

credential/manifest/validation.go
andresuribe87: This was mostly to make the `bare-return` happy.

credential/verification/verifiers.go
andresuribe87: Done

crypto/jwk_test.go
andresuribe87: Done

cryptosuite/jsonwebkey2020.go
andresuribe87: I like that, done!

example/usecase/employer_university_flow/pkg/util.go
andresuribe87: Using `did` would hid the package named `did`. I opted for `peer` instead, which is what the Generate method returns.

example/usecase/steel_thread/steel_thread.go
andresuribe87: Done.

decentralgabe (approved):
🚢

PR #248, opened 2022-11-04
Update to Presentation Exchange v2 and Latest CM

decentralgabe:


credential/manifest/builder_test.go
```diff
@@ -18,6 +18,12 @@ func TestCredentialManifestBuilder(t *testing.T) {
 
 	assert.False(t, builder.IsEmpty())
 
+	err = builder.SetName("name")
+	assert.NoError(t, err)
+
+	err = builder.SetDescription("description")
+	assert.NoError(t, err)
```
andresuribe87, line 25: Should this test verify that the correct values are set when the CM is built?

nitro-neal (approved):
:lit:

schema/known_schemas/pe-presentation-submission.json
```diff
@@ -28,7 +28,7 @@
           "$ref": "#/definitions/descriptor"
         },
         "format": {
-          "$ref": "https://identity.foundation/claim-format-registry/schemas/presentation-submission-claim-format-designations.json#/definitions/format"
+          "$ref": "http://identity.foundation/claim-format-registry/schemas/presentation-submission-claim-format-designations.json#/definitions/format"
```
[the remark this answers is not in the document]
decentralgabe: json schema breaks when you do 😢

credential/manifest/builder_test.go
decentralgabe: added

PR #249, opened 2022-11-05
Fix jwk usage

decentralgabe:
Support generating generic JWK objects. Add validation so that it's impossible to generate signers/verifiers with unsupported key types/algorithms.

### DIFF
--- a/crypto/jwk.go
+++ b/crypto/jwk.go
@@ -76,6 +76,24 @@ func JWKToPublicKeyJWK(key jwk.Key) (*PublicKeyJWK, error) {
 	return &pubKeyJWK, nil
 }
 
+// PublicKeyToJWK converts a public key to a JWK
+func PublicKeyToJWK(key crypto.PublicKey) (jwk.Key, error) {
+	switch key.(type) {
+	case rsa.PublicKey:
+		return jwkKeyFromRSAPublicKey(key.(rsa.PublicKey))
+	case ed25519.PublicKey:
+		return jwkKeyFromEd25519PublicKey(key.(ed25519.PublicKey))
+	case x25519.PublicKey:
+		return jwkKeyFromX25519PublicKey(key.(x25519.PublicKey))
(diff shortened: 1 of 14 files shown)


crypto/jwk.go
```diff
@@ -94,6 +112,24 @@ func PublicKeyToPublicKeyJWK(key crypto.PublicKey) (*PublicKeyJWK, error) {
...
+	case rsa.PrivateKey:
+		return jwkKeyFromRSAPrivateKey(key.(rsa.PrivateKey))
+	case ed25519.PrivateKey:
+		return jwkKeyFromEd25519PrivateKey(key.(ed25519.PrivateKey))
+	case x25519.PrivateKey:
+		return jwkKeyFromX25519PrivateKey(key.(x25519.PrivateKey))
+	case secp256k1.PrivateKey:
+		return jwkKeyFromSECP256k1PrivateKey(key.(secp256k1.PrivateKey))
+	case ecdsa.PrivateKey:
+		return jwkKeyFromECDSAPrivateKey(key.(ecdsa.PrivateKey))
```
andresuribe87, line 127:
```suggestion
	switch k := key.(type) {
	case rsa.PrivateKey:
		return jwkKeyFromRSAPrivateKey(k)
	case ed25519.PrivateKey:
		return jwkKeyFromEd25519PrivateKey(k)
	case x25519.PrivateKey:
		return jwkKeyFromX25519PrivateKey(k)
	case secp256k1.PrivateKey:
		return jwkKeyFromSECP256k1PrivateKey(k)
	case ecdsa.PrivateKey:
		return jwkKeyFromECDSAPrivateKey(k)
```

nitro-neal (approved):

andresuribe87 (commented):
small suggestion

crypto/jwk.go
decentralgabe: good idea, let me update this in the other one too

PR #255, opened 2022-11-23
Fix key type issue

andresuribe87:
This PR exposes the key type used when decoding the DIDKey. This is needed as otherwise the information would be lost when decoding.

This is a prerequisite to fixing https://github.com/TBD54566975/ssi-service/issues/181. 

Note that I had to update some examples of the tests, since the test vectors appear to have been incorrectly generated.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -25,7 +25,7 @@ linters-settings:
       - name: argument-limit
         severity: warning
         disabled: false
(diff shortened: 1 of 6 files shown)


decentralgabe (approved):
Looks good but may be cleaner as a new receiver method on DIDKey that is just `GetKeyType`

PR #257, opened 2022-11-28
Fix verification bug with JWT credentials

andresuribe87:
This fixes a bug where the presentations were throwing errors when the verifiable credentials contained in them were JWTs (instead of elements of type `VerifiableCredential`).

The test I added fails before the changes, and passes after the updates.

### DIFF
--- a/credential/exchange/verification.go
+++ b/credential/exchange/verification.go
@@ -2,6 +2,7 @@ package exchange
 
 import (
 	"fmt"
+	credutil "github.com/TBD54566975/ssi-sdk/credential/util"
 	"strings"
 
 	"github.com/TBD54566975/ssi-sdk/crypto"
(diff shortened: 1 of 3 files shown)


credential/exchange/verification.go
```diff
@@ -115,13 +116,19 @@ func VerifyPresentationSubmissionVP(def PresentationDefinition, vp credential.Ve
...
 			err := errors.Wrapf(err, "could not resolve claim from submission descriptor<%s> with path: %s",
 				submissionDescriptor.ID, submissionDescriptor.Path)
 			logrus.WithError(err).Error()
 			return err
 		}
+		submittedClaim, err := getClaimAsJSON(claim)
+		if err != nil {
+			err := errors.Wrapf(err, "getting claim as go-json: <%s>", claim)
+			logrus.WithError(err).Error()
+			return err
```
decentralgabe, line 130: `util.NewErrorf()`...

credential/exchange/verification.go
```diff
@@ -143,6 +150,28 @@ func VerifyPresentationSubmissionVP(def PresentationDefinition, vp credential.Ve
 	return nil
 }
 
+func getClaimAsJSON(claim interface{}) (map[string]interface{}, error) {
```
decentralgabe, line 153: should this be moved to the same pkg as `credutil.CredentialsFromInterface`?

nitro-neal (approved):
I did a quick check and it seems we are not using this in SSI-Service so we don't have to change anything there

credential/exchange/verification.go
andresuribe87: Keeping it here for consistency, and preventing scope creep. This should be addressed in https://github.com/TBD54566975/ssi-sdk/issues/258

PR #259, opened 2022-11-29
Standardize errors

### LINKED ISSUE #258: Standardize error treatment
**Is your feature request related to a problem? Please describe.**
An error that is thrown deep in the SDK will log multiple times, which is unnecessary and confusing for devs.

**Describe the solution you'd like**
It's reasonable to think that an SDK isn't responsible for logging. Instead, the system's using the SDK can decide what they want to do with an error. As such, I'm proposing we remove most logging of errors in the SDK, when the intention is to bubble up.

Additionally, we should include guidelines in the README on how we handle errors. 

**Describe alternatives you've considered**
An alternative is to log at the boundaries of the SDK. The problem is that there are often cases when an exported function calls another exported function. This makes it difficult to determine what constitutes the boundary. 

**Getting it done**
Some sort of massive scale refactoring. 

**Path to SIP**
Nope. 

**Additional context**
Example of current handling
```go
if err != nil {
	err := errors.Wrapf(err, "getting claim as go-json: <%s>", claim)
	logrus.WithError(err).Error()
	return err
}
```

Proposal of how to handle
```go
if err != nil {
	return errors.Wrapf(err, "getting claim as go-json: <%s>", claim)
}
```

decentralgabe:
Fixes #258

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -12,6 +12,7 @@ assignees: nitro-neal, decentralgabe
 
 **To Reproduce**
 *Steps to reproduce the behavior:*
+
 1. Go to '...'
 2. Click on '....'
 3. Scroll down to '....'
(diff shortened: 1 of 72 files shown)


decentralgabe:
also ran gofmt which changed a bunch!

decentralgabe:
Problem with a linter which I disabled as per https://github.com/golangci/golangci-lint/issues/3280

andresuribe87 (approved):

PR #265, opened 2022-12-09
Wasm bindings - a start

michaelneale:
This is the first pass add adding wasm binary (and an example web app that uses it from javascript) to ssi-sdk. 

thanks to @mistermoe for starting this and working out how it can work: https://github.com/TBD54566975/ssi-sdk-wasm-bindings-examples


magefile.go
```diff
@@ -31,7 +31,23 @@ const (
 // Build builds the library.
 func Build() error {
 	fmt.Println("Building...")
-	return sh.Run(Go, "build", "-tags", "jwx_es256k", "./...")
+	err := sh.Run(Go, "build", "-tags", "jwx_es256k", "./...")
+	if err != nil {
```
decentralgabe, line 35:
```suggestion
	if err := sh.Run(Go, "build", "-tags", "jwx_es256k", "./..."); err != nil {
```

magefile.go
```diff
@@ -31,7 +31,23 @@ const (
...
+		return err
+	}
+
+	return BuildWasm()
+
+}
+
+func BuildWasm() error {
```
decentralgabe, line 43:
```suggestion
	return BuildWASM()

}

func BuildWASM() error {
```

wasm/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+`webserver` is a sample webserver which serves up a sample js app from `static` which also contains the wasm bindings and wasm "binary".
+
+# status
+This just has some example usage of apis to start with.
+
+# building (from top level)
+
+`cd ssi-sdk`
+`GOOS=js GOARCH=wasm go build -tags jwx_es256k -o ./wasm/static/main.wasm ./wasm`
```
decentralgabe, line 14: should this be updated to `mage buildwasm`?

wasm/main.go
```diff
@@ -0,0 +1,62 @@
+//go:build js && wasm
+
+package main
+
+import (
+	"crypto/ed25519"
+	"encoding/base64"
+	"encoding/json"
```
decentralgabe, line 8: nit: we use `goccy/go-json` instead of `encoding/json`

wasm/main.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/TBD54566975/ssi-sdk/crypto"
+	"github.com/TBD54566975/ssi-sdk/did"
+)
+
+/*
+ * This is the glue to bind the functions into javascript so they can be called
+ */
+func main() {
+	done := make(chan struct{})
```
decentralgabe, line 19: curious what the difference is if this channel is removed?

wasm/main.go
```diff
@@ -0,0 +1,62 @@
...
+		}
+	}
+
+	return js.ValueOf("Unknown key type")
+}
+
+// 3. Returning a richer object, converting to json and then unmarshalling to make it a js object
+func makeDid(_ js.Value, args []js.Value) interface{} {
+
+	pubKey, _, _ := crypto.GenerateKeyByKeyType(crypto.Ed25519)
```
decentralgabe, line 52: should we pull out the errs here and below?

wasm/webserver/main.go
```diff
@@ -0,0 +1,24 @@
...
+
+	log.Println("Listening on http://localhost:3000/index.html")
+
+	server := &http.Server{
+		Addr:              ":3000",
+		ReadHeaderTimeout: 20 * time.Second,
+	}
+
+	err := server.ListenAndServe()
+	if err != nil {
```
decentralgabe, line 21:
```suggestion
	if err := server.ListenAndServe(); err != nil {
```

decentralgabe (commented):
few small comments - awesome to get this going

decentralgabe:
Screenshot 2022-12-09 at 10 10 13 AM
two vulns

wasm/main.go
andresuribe87: IIRC, you need to keep the process alive.

wasm/main.go
```diff
@@ -0,0 +1,62 @@
...
+	keyType := args[0].String()
+	for _, k := range crypto.GetSupportedKeyTypes() {
+		if string(k) == keyType {
+			publicKey, _, _ := crypto.GenerateKeyByKeyType(k)
+			pubKeyBytes, _ := crypto.PubKeyToBytes(publicKey)
+			return js.ValueOf(base64.StdEncoding.EncodeToString(pubKeyBytes))
+		}
+	}
+
+	return js.ValueOf("Unknown key type")
```
andresuribe87, line 46: Consider using `IsSupportedKeyType` and returning early to reduce nesting. 
```suggestion
        kt := crypto.KeyType(keyType)
	if !crypto.IsSupportedKeyType(kt) {
	  return js.ValueOf("Unknown key type")
	}
	publicKey, _, _ := crypto.GenerateKeyByKeyType(kt)
	pubKeyBytes, _ := crypto.PubKeyToBytes(publicKey)
	return js.ValueOf(base64.StdEncoding.EncodeToString(pubKeyBytes))
```

wasm/static/index.html
```diff
@@ -0,0 +1,31 @@
+<html>
+
+<head>
+  <meta charset="utf-8" />
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/ace/1.13.1/ace.js" integrity="sha512-IQmiIneKUJhTJElpHOlsrb3jpF7r54AzhCTi7BTDLiBVg0f7mrEqWVCmOeoqKv5hDdyf3rbbxBUgYf4u3O/QcQ==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
```
andresuribe87, line 5: Curious why this is needed?

andorsk:
@michaelneale i've done some work on this actually on my side, but never merged it/sent a PR. Would be happy to sync about it and show some of the stuff I did if that helps.

michaelneale:
@andorsk sure - could open it in a branch if you like or a gist?

andorsk:
@michaelneale  yea...i need to do some merging with/cleanup to make it work directly with the ssi-sdk ( it's over benri's sdk, but yea..I can either send a PR to this or a new PR. Not sure if GIST makes sense as there is a few files.

Here's an example of a VC I generated using WASM + ssk-sdk 

image

Couple things I recommend paying attention to:

1. I ended up namespacing my commands. 
2. For mage, I ended up abstracting some of the commands to work a little better with the multiple combos that exist.
3. There's one major issue which I need to investigate:
   - Restarting the connection if it fails and keeping a connection open. It's probably some singleton like format. I tried some polling on my side, but it seems to keep loading the artifact entirely each poll so something is wrong there.

michaelneale:
interesting - I did suspect error handling would need a lot of work.

mistermoe:
@andorsk +1 to namespacing commands

did some digging around about how to handle error handling. it's unfortunately weird. found [this](https://stackoverflow.com/questions/67437284/how-to-throw-js-error-from-go-web-assembly) stackoverflow post which sorta highlights how error handling is weird and  lays out 2 approaches (1 of which is kinda yikes)

mistermoe:
eventually, might be nice to have a test suite for the bindings. can use any js testing framework and [karma](https://karma-runner.github.io/latest/index.html) for running tests in a headless browser

wasm/static/index.html
mistermoe: @andresuribe87 it was just used to make the did doc json output look fancy. can def remove it

wasm/main.go
```diff
@@ -0,0 +1,62 @@
...
+
+// 1. Simplest function - note we wrap things with js.ValueOf (if a primitive you don't technically need to)
+func sayHello(_ js.Value, args []js.Value) interface{} {
+	return js.ValueOf("Hello from golang via wasm!")
+}
+
+// 2. Calling a ssi-sdk function directly - but returning a plain old string
+func generateKey(_ js.Value, args []js.Value) interface{} {
+
+	keyType := args[0].String()
```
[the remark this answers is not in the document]
mistermoe: for sure @andresuribe87  this method will currently blow up. the most i'd do in this PR though is add a TODO above this line saying that arg length needs to be checked and handled appropriate (aka throw error if no args provided). Then, a general pattern can be figured out and applied for error handling in another PR

andorsk:
@mistermoe a couple things here: 

1. for error handling, yes. it's a giant nightmare. for a couple other reasons. 1. It's hard to throw an error
2. When you start the service, if the service panics for some it crashes the service. And unless you build a mechanic to catch the panic (i.e `recover`) your service dies. 
3. re: some test suite, yes, but I will say I've done it on my side, and I thought I had it working but it's currently broken ( some issue with build tags ), so now I'm not so sure.  It's also annoying b/c you have to keep converting things to js.Value.

wasm/main.go
mistermoe: ✅

michaelneale:
thanks for comments. FYI the `main.go` is mainly as a first example to get things going, but I guess we need to formalise how to expose these apis with the examples there.

michaelneale:
@andorsk are you happy for me to continue plugging along on this to get wasm in SSI-SDK and then we can iterate on it? or we need a different approach?

andorsk:
@michaelneale I think continuing to plug away on it is important, but my suggestion here is: let's get two functions with a dependency really stable. With the mechanics to do it and a way to move from 1 -> N functions easily. Then we can build the rest of the functions out.

michaelneale:
@andorsk great thoughts - agree- will keep plugging away with hopefully something simple end to end that can be added to. 
I am a little confused about namespacing functions thought, is that to avoid polluting the global namespace on the js side?

andorsk:
exactly. I just hated the idea of conflicting with the global namespace. I'm not as great with JS stuff as I am with Go and so forth, and particularly with scoping I get thrown around in a dizzy. There might be a better way to do it. ( i.e a WASM object ).

michaelneale:
hey @andorsk tidied some things up, added in a did resolver to try out: 

`console.log(resolveDid("did:peer:0z6MkpTHR8VNsBxYAAWHut2Geadd9jSwuBV8xRoAnwWsdvktH"));`

And some retry logic and more (still no tests or good error handling). 

Also, and I am ashamed to say this, that function was written mostly by AI.

wasm/main.go
```diff
@@ -73,3 +74,30 @@ func makeDid(_ js.Value, args []js.Value) interface{} {
 	return js.ValueOf(resultObj)
 
 }
+
+func resolveDid(_ js.Value, args []js.Value) interface{} {
```
nitro-neal, line 78: do we have a way to distinguish in the JS if this was an error vs the return object?

nitro-neal (approved):

wasm/main.go
andorsk: that's a great question. I've had a similar question and I've not been able to determine a way aside from serializing a string with the word "error". Maybe the right way is to build a wrapper around the objects such as the following:

```go
type ObjectType string

TypeError ObjectType = "error"
TypeObject ObjectType = "object"

type ObjectWrapper  {
     ObjectType ObjectType `json:"type"`
     Object interface{} `json:"object"`
}
```

or something. There's no "standard" way though, AFAIK.

wasm/main.go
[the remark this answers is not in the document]
andorsk: agreed here. Here was how I did it, but I definitely would want to revisit it. I didn't feel it was robust enough: 

```go
func checkArgs(actual []js.Value, args ...string) error {
	if len(actual) < len(args) {
		return errors.New(fmt.Sprintf("not enough arguments. Need %v", args))
	}
	return nil
}
```
then call it later: 

```
err := checkArgs(args, "id")
if err != nil {
   return err.Error()
}
```

re: blow up: if this blows up, it doesn't recover. Which is a big issue. 

So I suggest we have a recover mechanic in place before getting too deep into actually building out the methods.

wasm/main.go
andorsk: Another thing to think about with WASM is how it interfaces with storage. It's not quite clear cut. I.e you generate the public key in WASM, but you'll now how to figure out a way to transfer and serialize the key to the application handling the storage. That transfer always felt a little "iffy" to me. I.e you've now sent your private key to your application, and that may be a security risk.

andorsk (commented):
My initial thoughts here are:

It's a great starting point! But we need to make sure to iron out the "handling" more before going onto actually building out useful things. Such as: 

1. Error Handling
2. Recovery
3. Exiting the service
4. Argument handling
5. Namespacing. 
6. Serialization and APIs of keys and other secure things. 
7. Test suite

wasm/main.go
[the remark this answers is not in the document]
michaelneale: @andorsk wouldn't the main function trap this and restart?

wasm/main.go
michaelneale: @andorsk I am not that worried about that at the moment: I like the idea of using WASM for pure computation (ie functions without side effects and no state outside its bounds) were possible especially with ssi-ssd (people can correct me if I am wrong). I know you could build your whole app and state in wasm, but I don't think that is needed here is it?

wasm/main.go
mistermoe: @nitro-neal  yeah there's no easy way to throw a JS exception from web assembly to support something like

```js
try {
  resolveDid("did:janky:alice")
} catch (e) {
  console.log(error);
}
```

WASM execution happens within an isolated sandbox in a separate execution environment that cannot directly access the JS stack. So anything thrown from WASM gets caught by the WASM runtime and handled as if the WASM code panicked.

@michaelneale your top level `error` property could definitely work. Another option that may feel a bit more like catching an exception would be to return a `Promise` from WASM that is `resolve`d or `reject`ed. Using `await` semantics you'd be able to try/catch the returned promise, e.g.

```js
try {
  await resolveDid("did:janky:alice")
} catch (e) {
  console.log(error);
}
```

the underlying Go code would look like this:
```go
func resolveDid(_ js.Value, args []js.Value) interface{} {
    handler := js.FuncOf(func(this js.Value, args []js.Value) interface{} {
        resolve := args[0]
        reject := args[1]
    
        go func() {
            didString := args[0].String()
            resolvers := []did.Resolution{did.KeyResolver{}, did.WebResolver{}, did.PKHResolver{}, did.PeerResolver{}}
            resolver, err := did.NewResolver(resolvers...)
            
            if err != nil {
                // err should be an instance of `error`, eg `errors.New("some error")`
                errorConstructor := js.Global().Get("Error")
                errorObject := errorConstructor.New(err.Error())
                reject.Invoke(errorObject)
            }
    
            doc, err := resolver.Resolve(didString)
            if err != nil {
                errorConstructor := js.Global().Get("Error")
                errorObject := errorConstructor.New(err.Error())
                reject.Invoke(errorObject)
            }
    
            resultBytes, err := json.Marshal(doc)
            if err != nil {
                errorConstructor := js.Global().Get("Error")
                errorObject := errorConstructor.New(err.Error())
                reject.Invoke(errorObject)
            }
    
            var resultObj map[string]interface{}
            err = json.Unmarshal(resultBytes, &resultObj)
            if err != nil {
                errorConstructor := js.Global().Get("Error")
                errorObject := errorConstructor.New(err.Error())
                reject.Invoke(errorObject)
            }
    
            resolve.Invoke(js.ValueOf(resultObj))
        }()
    
        return nil
    })
    
    promiseConstructor := js.Global().Get("Promise")
    return promiseConstructor.New(handler)
}
```

the go is def a bit uglier but i bet we can teach ChatGPT to return functions with this kind of wrapper.


random unnecessary info:
the "uneasy" way to throw exceptions from WASM into the JS runtime i think would require us to provide the assembler with a hint. so something like:

```go
// throw stub in our go code
func Throw(exception string, message string)
```

```
// assembler hint in ssi-sdk_js.s

TEXT ·Throw(SB), NOSPLIT, $0
  CallImport
  RET
```

```js
// manually added to wasm_exec.js into `importObject.go` object
// this object already exists in `wasm_exec.js`, just adding here for sake of example
this.importObject = {
// `go` already exists in `wasm_exec.js`. just adding here for sake of example
  go: {
    // this is what we would need to add ourselves
    // func Throw(exception string, message string)
    'ssi-sdk.Throw': (sp) => {
      const exception = loadString(sp + 8)
      const message = loadString(sp + 24)
      const throwable = globalThis[exception](message)
      throw throwable
    }
  }
}
```

andresuribe87: Heavy +1 to returning a promise. DID resolution will typically involve making a network request, which reinforces the fact that it should be an awaitable. 

In fact, I can see the [Resolve](https://github.com/TBD54566975/ssi-sdk/blob/92be548d54be11b3740b95bb5c568958e5979de0/did/resolver.go#L16) method evolving to become more go idiomatic and accepting a `context.Context` object as the first parameter. This object would carry cancellation signals.

wasm/main.go
andorsk: @michaelneale fair points. I get using WASM for pure computation, and think a stateless app is probably fair.  One question though here: moving keys back and forth between local storage to the WASM application. I don't have an answer here, but does shuffling the key around between WASM and JS runtime render any security issues?

i.e 

```js
var pk = JSON.parse(CreatePK()) // generates PK. Now key is JS Object. You could store this somewhere after.
doX(JSON.stringify(pk)) // PK is serialized and sent. 
```

wasm/main.go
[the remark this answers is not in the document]
andorsk: if panic, it would exit main function and it wouldn't restart unless you handled it. At least, my experience.

andorsk:
@michaelneale Thanks for being patient with some of my comments. Re: some of my comments around better handling, I think it would be reasonable to address them in a future PR. 

I know there was a lot there, and it probably makes sense to address them after, but note that they are pretty important to have a plan around before expanding features.

This work was really great though! Thanks for putting in the effort to get this into a reality.

michaelneale:
@andorsk no worries at all - has been most helpful! 

@andresuribe87 @nitro-neal do you want to take this one over? can merge it as is, or add in Moe's changes for error handling? (either as part of this or a future PR) so that we can expose things as needed?

andresuribe87:
I can take over! I would suggest merging as is and then fixing as we go.

michaelneale:
done!

PR #269, opened 2022-12-13
Locally load schemas

### LINKED ISSUE #134: Have the ability to cache / load local copies of referenced schemas

decentralgabe:
Fixes #134 

Don't think this will work for schemas that are not pre-defined. Meaning, if we're in local mode and try to load a http/s resource that's not defined it'll break. This should be fine...

It is 'turned on' by default with an `init()` block which isn't ideal. Could use some ideas to make this better.


decentralgabe:
Tested by turning off wifi and running the tests 😉

schema/known_schemas/pe-submission-claim-format-designations.json
```diff
@@ -2,17 +2,10 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
```
nitro-neal, line 2: did everything else get deleted but not this file?

schema/jsonschema.go
```diff
@@ -14,21 +13,17 @@ import (
...
-	knownSchemas embed.FS
-)
-
 func init() {
 	httploader.Client = &http.Client{
 		Timeout: time.Second * 10,
 	}
 }
 
+const (
```
nitro-neal, line 22: put the const above init

nitro-neal (approved):
Good job! We the way we had it before was so janky

schema/jsonschema.go
```diff
@@ -14,21 +13,17 @@ import (
...
+	// defaultSchemaURL is a placeholder that's needed to load any schema
+	defaultSchemaURL = "schema.json"
```
andresuribe87, line 24: Not sure I follow why this is needed, can you elaborate?

schema/loader.go
```diff
@@ -0,0 +1,144 @@
...
+		"identity.foundation/credential-manifest/schemas/credential-application.json":                              CredentialApplicationSchema,
+		"identity.foundation/credential-manifest/schemas/credential-response.json":                                 CredentialResponseSchema,
+		"identity.foundation/credential-manifest/schemas/output-descriptors.json":                                  OutputDescriptorsSchema,
+		"identity.foundation/wallet-rendering/schemas/display-mapping-object.json":                                 DisplayMappingObjectSchema,
+		"identity.foundation/wallet-rendering/schemas/entity-styles.json":                                          EntityStylesSchema,
+		"identity.foundation/wallet-rendering/schemas/labeled-display-mapping-object.json":                         LabeledDisplayMappingObjectSchema,
+		"w3c-ccg.github.io/vc-json-schemas/credential-schema-2.0":                                                  VerifiableCredentialJSONSchemaSchema,
+	}
+
+	localSchemas := make(map[string]string)
```
andresuribe87, line 119:
```suggestion
	localSchemas := make(map[string]string, len(localFiles))
```

schema/loader.go
```diff
@@ -0,0 +1,144 @@
...
+
+	localLoader := NewLocalLoader()
+	if err := localLoader.AddLocalLoad(localSchemas); err != nil {
+		return errors.Wrap(err, "could not load local schemas")
+	}
+	return nil
+}
+
+// DisableLocalLoad disables the ability to load schemas locally, clearing current locally loaded schemas
+func DisableLocalLoad() {
```
andresuribe87, line 136: Is this being used?

schema/loader.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+		return io.NopCloser(strings.NewReader(schema)), nil
+	}
+	jsonschema.Loaders["http"] = func(url string) (io.ReadCloser, error) {
+		schema, ok := localSchemas[strings.TrimPrefix(url, "http://")]
+		if !ok {
+			return nil, fmt.Errorf("%q not found", url)
+		}
+		return io.NopCloser(strings.NewReader(schema)), nil
+	}
```
andresuribe87, line 84: I have a couple of suggestions for your consideration:
* Rework `LocalLoader` into `LocalLoaderFactory`.
* Remove the `jsonschema.Loader` directives from this file, and the `init` functions (let users of the SDK do it themselves).
* Provide a method for creating loaders, something like what's below. 

```go
type Loader func(url string) (io.ReadCloser, error)

func (l *LocalLoaderFactory) ProtocolLoader(protocol string) Loader {
  return func(url string) (io.ReadCloser, error) {
		schema, ok := localSchemas[strings.TrimPrefix(url, protocol + "://")]
		if !ok {
			return nil, fmt.Errorf("%q not found", url)
		}
		return io.NopCloser(strings.NewReader(schema)), nil
	}
}
```

Once that is done, then user's of the SDK can decide where to override the `jsonschema` loader. 

```go
// Somewhere in the ssi-service

loaderFactory := schema.NewLocalLoaderFactory()
jsonschema.Loader["http"] = loaderFactory.ProtocolLoader("http")
jsonschema.Loader["https"] = loaderFactory.ProtocolLoader("https")
```

Thoughts?

dwn/model.go
```diff
@@ -46,29 +46,8 @@ func (msg *DWNMessage) IsValid() error {
 		return errors.New("manifest is empty")
 	}
 
-	// validate against json schema
-	if err := IsValidDWNMessage(*msg); err != nil {
-		return errors.Wrap(err, "dwn message failed json schema validation")
-	}
+	// TODO: validate the message against a json schema
```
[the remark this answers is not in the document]
decentralgabe: the method was commented out and would need to be significantly reworked before using it again.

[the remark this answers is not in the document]
decentralgabe: I am also not sure if this file is being used @nitro-neal is it dead code?

schema/jsonschema.go
decentralgabe: the library requires you to provide a url for the schema and we don't really care what this is but it needs to be a placeholder https://github.com/santhosh-tekuri/jsonschema/blob/master/compiler.go#L65

schema/known_schemas/pe-submission-claim-format-designations.json
```diff
@@ -2,17 +2,10 @@
...
   "title": "Presentation Submission Claim Format Designations",
   "type": "object",
-  "additionalProperties": false,
-  "properties": {
+  "definitions": {
```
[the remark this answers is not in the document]
decentralgabe: I didn't delete most schemas. I deleted the `-deref.json` files. This schema actually had an update which I pulled in. cc: @nitro-neal

schema/loader.go
andresuribe87: How about defining the TestMain (as described in this https://pkg.go.dev/testing#hdr-Main)? Within that function we can patch the jsonschema lib.

schema/loader.go
```diff
@@ -0,0 +1,154 @@
...
+	var schemas []string
+	for schemaURI := range cl.schemas {
+		schemas = append(schemas, schemaURI)
+	}
+	return schemas, nil
+}
+
+// AddCachedSchema adds a schema to be cached
+func (cl *CachingLoader) AddCachedSchema(schemaURI, schema string) error {
```
decentralgabe, line 94: adjusted this to be a caching loader with a test. no more init.

schema/loader.go
```diff
@@ -0,0 +1,154 @@
...
+
+// CachingLoader is a struct that holds local schemas
+type CachingLoader struct {
+	schemas map[string]string
+}
+
+// NewCachingLoader returns a new CachingLoader that enables the ability to cache http and https schemas
+func NewCachingLoader() CachingLoader {
+	localSchemas := make(map[string]string)
+	jsonschema.Loaders["https"] = func(url string) (io.ReadCloser, error) {
```
andresuribe87, line 64: I find it odd that a constructor has side effects. What's the rationale of patching `jsonschema` within the constructor?

schema/loader.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	return CachingLoader{schemas: localSchemas}
+}
+
+// GetCachedSchemas returns an array of cached schema URIs
+func (cl *CachingLoader) GetCachedSchemas() ([]string, error) {
+	if cl.schemas == nil {
+		return nil, errors.New("caching loader is not instantiated")
+	}
+	var schemas []string
```
andresuribe87, line 86:
```suggestion
	schemas := make([]string, len(cl.schemas))
```

schema/loader.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	for schemaURI, schema := range schemas {
+		if err := cl.AddCachedSchema(schemaURI, schema); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// LoadSchema loads a schema from the embedded filesystem
```
andresuribe87, line 118:
```suggestion
// LoadSchema loads a schema from the embedded filesystem and returns it's contents.
```

schema/loader_test.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"testing"
+
+	"github.com/TBD54566975/ssi-sdk/util"
+	"github.com/jarcoal/httpmock"
+	"github.com/santhosh-tekuri/jsonschema/v5"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCachingLoader(t *testing.T) {
```
andresuribe87, line 12: Consider splitting this large test into multiple smaller unit tests that exercise a single behavior.

schema/loader.go
```diff
@@ -0,0 +1,154 @@
...
+// CachingLoader is a struct that holds local schemas
+type CachingLoader struct {
+	schemas map[string]string
+}
+
+// NewCachingLoader returns a new CachingLoader that enables the ability to cache http and https schemas
+func NewCachingLoader() CachingLoader {
+	localSchemas := make(map[string]string)
+	jsonschema.Loaders["https"] = func(url string) (io.ReadCloser, error) {
+		schema, ok := localSchemas[strings.TrimPrefix(url, "https://")]
```
andresuribe87, line 65: This might not be safe for concurrent use (and cause a panic), depending on how users of the SDK exercise it. 

A goroutine might be reading from the map, while another one writes to the map by calling `AddCachedSchema`.

Consider writing once to the map only, in the constructor. Concurrent reads on a map are safe.

schema/jsonschema.go
decentralgabe: done

schema/loader.go
decentralgabe: most, if not all, schemas will be loaded by http/s. having users configure additional loaders leaks details of the library that they shouldn't know or need to care about. this loader will work fine for the 99%

schema/loader.go
decentralgabe: this constructor should only be called once - and I don't imagine callers will be further configuring the loaders...if they do then yes they may have concurrency issues. is this worth caring about?

schema/loader_test.go
decentralgabe: split into 2

dwn/model.go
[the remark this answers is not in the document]
nitro-neal: yup dead code, no dwn stuff should live in ssi-service or sdk

schema/loader.go
```diff
@@ -0,0 +1,165 @@
...
+	return &cl, nil
+}
+
+// EnableHTTPCache enables caching of http and https schemas
+func (cl *CachingLoader) EnableHTTPCache() {
+	jsonschema.Loaders["http"] = cl.cachingLoaderForProtocol("http")
+	jsonschema.Loaders["https"] = cl.cachingLoaderForProtocol("https")
+}
+
+func (cl *CachingLoader) cachingLoaderForProtocol(protocol string) func(url string) (io.ReadCloser, error) {
```
decentralgabe, line 81: updated to lock on map writes.

schema/loader.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+
+		// load from the internet
+		loaded, err := httploader.Load(url)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to load schema from %s", protocol)
+		}
+
+		// read the contents and cache and prevent future lookups
+		contents, err := io.ReadAll(loaded)
```
decentralgabe, line 107: added cache on load from http

dwn/model.go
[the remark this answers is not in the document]
decentralgabe: removed

schema/loader.go
```diff
@@ -0,0 +1,165 @@
...
+		// fallback lookup if it's stored with the fully qualified url
+		schema, ok = cl.schemas[url]
+		if ok {
+			return io.NopCloser(strings.NewReader(schema)), nil
+		}
+
+		// load from the internet
+		loaded, err := httploader.Load(url)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to load schema from %s", protocol)
```
andresuribe87, line 103: So that the error reads better
```suggestion
			return nil, errors.Wrapf(err, "loading schema from %s", protocol)
```

schema/loader.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+
+		// read the contents and cache and prevent future lookups
+		contents, err := io.ReadAll(loaded)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to read schema from %s", protocol)
```
andresuribe87, line 109:
```suggestion
			return nil, errors.Wrapf(err, "reading all %s", protocol)
```

schema/loader.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// CachingLoader is a struct that holds local schemas
+type CachingLoader struct {
+	schemas map[string]string
+}
+
+// NewCachingLoader returns a new CachingLoader that enables the ability to cache http and https schemas
+func NewCachingLoader(schemas map[string]string) (*CachingLoader, error) {
+	cl := CachingLoader{schemas: make(map[string]string)}
```
andresuribe87, line 65:
```suggestion
	cl := CachingLoader{schemas: make(map[string]string, len(schemas))}
```

schema/loader.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// EnableHTTPCache enables caching of http and https schemas
+func (cl *CachingLoader) EnableHTTPCache() {
+	jsonschema.Loaders["http"] = cl.cachingLoaderForProtocol("http")
+	jsonschema.Loaders["https"] = cl.cachingLoaderForProtocol("https")
+}
+
+func (cl *CachingLoader) cachingLoaderForProtocol(protocol string) func(url string) (io.ReadCloser, error) {
+	var m sync.Mutex
```
andresuribe87, line 82: Did you consider using a `sync.RWMutex` in order to lock more granularly? It could increase read throughput (though that's a small optimization).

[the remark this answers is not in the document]
decentralgabe: yes but isn't that fine, since they're different protocols?

schema/loader.go
```diff
@@ -0,0 +1,159 @@
...
+
+// CachingLoader is a struct that holds local schemas
+type CachingLoader struct {
+	schemas sync.Map
+}
+
+// NewCachingLoader returns a new CachingLoader that enables the ability to cache http and https schemas
+func NewCachingLoader(schemas map[string]string) (*CachingLoader, error) {
+	// make sure only one process can write to the map at a time
+	cl := CachingLoader{schemas: sync.Map{}}
```
andresuribe87, line 66: nit: https://github.com/uber-go/guide/blob/master/style.md#omit-zero-value-fields-in-structs

```suggestion
	cl := CachingLoader{}
```

schema/loader.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// CachingLoader is a struct that holds local schemas
+type CachingLoader struct {
+	schemas sync.Map
+}
+
+// NewCachingLoader returns a new CachingLoader that enables the ability to cache http and https schemas
+func NewCachingLoader(schemas map[string]string) (*CachingLoader, error) {
+	// make sure only one process can write to the map at a time
```
andresuribe87, line 65: nit: move this to be a comment for the `schemas` function.

andresuribe87 (approved):
This is awesome, I'm so excited about this change. Thanks for tackling!

Issue #261, opened 2022-12-02
[Idea] Add fuzz testing


andresuribe87:
**Is your feature request related to a problem? Please describe.**
The SDK can be hardenned further by incorporating fuzz tests in many of the functions.

**Describe the solution you'd like**
Fuzz tests added in the `did` package. A good candidate is the `key` DID method. A fuzz test could test the property that "Generate and then resolve returns a document with the same ID". 

**Describe alternatives you've considered**
N/A

**Getting it done**
Looking for contibutors :) 

**Path to SIP**
No, this is a simple test to be added for hardening. 

**Additional context**
See this doc for more details on how to do fuzz testing: https://go.dev/security/fuzz/

PR #270, opened 2022-12-13
[WIP] #261: Add fuzz test

td-bn:
## Overview
WIP PR for [issue#261](https://github.com/TBD54566975/ssi-sdk/issues/261)

## Description
Adds:
- fuzz test for "Create and then resolve returns a document with the same ID".
- Unit tests for "Generate and then resolve returns a document with the same ID".

## How Has This Been Tested?
With commands:
`mage fuzz`
`mage test`

### DIFF
--- a/did/key_fuzz_test.go
+++ b/did/key_fuzz_test.go
@@ -1,0 +1,56 @@
+package did
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var mockPubKeys = []string{
+	"b9c5714089478a327f09197987f16f9e5d936e8a", "5f246d7d19aa612d6718d27c1da1ee66859586b0", "7d2d43e63666f45b40316b44212325625dbaeb40", "1c1f02f1640e52b313f2d504b3c0c7ee8ad61108", "69c5888ecd21287fbdac5a43d1558bf73c51e38b",
+}
+
+func FuzzCreateAndDecode(f *testing.F) {
+	keytypes := GetSupportedDIDKeyTypes()
+	ktLen := len(keytypes)
+
+	for i, pk := range mockPubKeys {
+		f.Add(i, []byte(pk))
+	}
+	f.Fuzz(func(t *testing.T, ktSeed int, pubKey []byte) {
+		kt := keytypes[(ktSeed%ktLen+ktLen)%ktLen]
+
+		didKey, err := CreateDIDKey(kt, pubKey)
+		assert.NoError(t, err)
+
+		recvPubKey, _, _, err := didKey.Decode()
+		assert.NoError(t, err)
+		assert.Equal(t, pubKey, recvPubKey)
+	})
+}
+
+func FuzzCreateAndResolve(f *testing.F) {
+	keytypes := GetSupportedDIDKeyTypes()
+	ktLen := len(keytypes)
+
+	resolvers := []Resolution{KeyResolver{}, WebResolver{}, PKHResolver{}, PeerResolver{}}
+	resolver, _ := NewResolver(resolvers...)
+
+	for i, pk := range mockPubKeys {
+		f.Add(i, []byte(pk))
+	}
+
+	f.Fuzz(func(t *testing.T, ktSeed int, pubKey []byte) {
+		kt := keytypes[(ktSeed%ktLen+ktLen)%ktLen]
+
+		didKey, err := CreateDIDKey(kt, pubKey)
+		assert.NoError(t, err)
+
+		doc, err := resolver.Resolve(didKey.String())
+		if err != nil {
+			t.Skip()
+		}
+		assert.NotEmpty(t, doc)
+		assert.Equal(t, didKey.String(), doc.DIDDocument.ID)
+	})
+}
--- a/did/key_test.go
+++ b/did/key_test.go
@@ -196,6 +196,37 @@ func TestExpand(t *testing.T) {
 	})
 }
 
+func TestGenerateAndDecode(t *testing.T) {
+	for _, kt := range GetSupportedDIDKeyTypes() {
+		privKey, didKey, err := GenerateDIDKey(kt)
+		assert.NotEmpty(t, privKey)
+		assert.NoError(t, err)
+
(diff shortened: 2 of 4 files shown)


td-bn:
First time working with `go` 🙏 . With this PR I'm trying to capture expectations and get early feedback.

Looking for pointers to improve and extend this PR - specifically  (but not limited to):
- better way to fuzz test these properties
- changes required to capture such tests in a mage command

did/key_fuzz_test.go
```diff
@@ -0,0 +1,43 @@
+package did
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func FuzzGenerateAndDecode(f *testing.F) {
+	keytypes := GetSupportedDIDKeyTypes()
```
andresuribe87, line 10: First off, this is awesome; thanks for taking a stab at it!

That said, I realize now that these tests are likely better suited as unit test that go over all keytypes.

Fuzz testing works best when there is a function that can accept arbitrary random input from the fuzz arguments. 

I realize that this isn't the best case for `GenerateDIDKey`. So my suggestion would be to:
* Move the existing tests you wrote to unit tests which go over all keytypes.
* Do similar fuzz tests as what you have, but calling `CreateDIDKey` instead. You can pass in the the `publicKey []byte` as a fuzz argument. For key type, you can do the same thing you're currently doing. 
  * Make sure to verify that the `publicKey` passed in is the same as what results after calling `Decode` :)

andresuribe87 (commented):
Regarding mage, I'd recommend adding a separate command for fuzzing, and making sure you include the flag `-fuzztime`. A sensible default can be 10 seconds. 

After you add the mage command, we can include it as part of the CI pipeline. Let me know if that all makes sense and thanks!

did/key_fuzz_test.go
```diff
@@ -6,37 +6,51 @@ import (
...
-	f.Fuzz(func(t *testing.T, a uint) {
-		kt := keytypes[a%uint(len(keytypes))]
 
-		privKey, didKey, err := GenerateDIDKey(kt)
-		assert.NotEmpty(t, privKey)
+	pks := []string{"b9c5714089478a327f09197987f16f9e5d936e8a", "5f246d7d19aa612d6718d27c1da1ee66859586b0", "7d2d43e63666f45b40316b44212325625dbaeb40", "1c1f02f1640e52b313f2d504b3c0c7ee8ad61108", "69c5888ecd21287fbdac5a43d1558bf73c51e38b"}
+	for i, pk := range pks {
+		f.Add(uint(i), []byte(pk))
+	}
+	f.Fuzz(func(t *testing.T, a uint, pubKey []byte) {
```
decentralgabe, line 16: what is `a` can it be a more meaningful variable name?

did/key_fuzz_test.go
```diff
@@ -6,37 +6,51 @@ import (
...
+func FuzzCreateAndDecode(f *testing.F) {
 	keytypes := GetSupportedDIDKeyTypes()
-	f.Fuzz(func(t *testing.T, a uint) {
-		kt := keytypes[a%uint(len(keytypes))]
 
-		privKey, didKey, err := GenerateDIDKey(kt)
-		assert.NotEmpty(t, privKey)
+	pks := []string{"b9c5714089478a327f09197987f16f9e5d936e8a", "5f246d7d19aa612d6718d27c1da1ee66859586b0", "7d2d43e63666f45b40316b44212325625dbaeb40", "1c1f02f1640e52b313f2d504b3c0c7ee8ad61108", "69c5888ecd21287fbdac5a43d1558bf73c51e38b"}
+	for i, pk := range pks {
+		f.Add(uint(i), []byte(pk))
```
decentralgabe, line 14: why cast to a uint?

did/key_fuzz_test.go
```diff
@@ -6,37 +6,51 @@ import (
...
+	}
+	f.Fuzz(func(t *testing.T, a uint, pubKey []byte) {
+		kt := keytypes[a%uint(len(keytypes))]
 
+		didKey, err := CreateDIDKey(kt, pubKey)
+		if err != nil {
+			t.Skip()
+		}
 		expectedLdKeyType, _ := KeyTypeToLDKeyType(kt)
```
decentralgabe, line 23: let's err check here

and nit: `expectedLDKeyType`

did/key_test.go
```diff
@@ -196,6 +196,36 @@ func TestExpand(t *testing.T) {
 	})
 }
 
+func TestGenerateAndDecode(t *testing.T) {
+	for _, kt := range GetSupportedDIDKeyTypes() {
+		privKey, didKey, err := GenerateDIDKey(kt)
+		assert.NotEmpty(t, privKey)
+
+		expectedLdKeyType, _ := KeyTypeToLDKeyType(kt)
```
decentralgabe, line 204:
```suggestion
		expectedLLKeyType, _ := KeyTypeToLDKeyType(kt)
```

magefile.go
```diff
@@ -64,6 +68,92 @@ func runTests(extraTestArgs ...string) error {
...
+			if err != nil {
+				return err
+			}
+		}
+
+	}
+	return nil
+}
+
+var FuzzFuncNameRegex = regexp.MustCompile(`\bFuzz\w+`)
```
decentralgabe, line 102: does this need to be exported?

magefile.go
```diff
@@ -64,6 +68,92 @@ func runTests(extraTestArgs ...string) error {
...
+
+// Searches the given directory recursively for Go files containing the word "Fuzz" in their function names.
+// Returns the names of the functions that match the search criteria.
+func getFuzzTests(dir string) ([]string, error) {
+	results := []string{}
```
decentralgabe, line 107:
```suggestion
	var results []string
```

andresuribe87 (commented):
Thanks for doing this! One small comment but looks great.

andresuribe87:
Ohhh I forgot - it would make sense to include this in the github CI. Take a look at the `.github/workflows/ci.yml` file.

andresuribe87 (approved):

Issue #272, opened 2023-01-17
Uptake VC Media Types


decentralgabe:
As per https://github.com/w3c/vc-data-model/pull/1000
Support media types for different representations of credentials

decentralgabe:
consider use of https://transmute-industries.github.io/vc-jws/ in place of vc-jwt

PR #273, opened 2023-01-25
Added signing utilities for vc-jws, with the appropriate content type.

andresuribe87:
This fixes https://github.com/TBD54566975/ssi-sdk/issues/272

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,6 +1,8 @@
 # See https://golangci-lint.run/usage/configuration/ for reference.
 run:
   concurrency: 16
+  build-tags:
+    - jwx_es256k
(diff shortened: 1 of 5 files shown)


decentralgabe (approved):
I like this, but I'm hesitant to start using it given it hasn't seen any consensus or adoption (yet) in the working group. I wonder if we could add a go build tag for "experimental" or similar.

andresuribe87:
> I wonder if we could add a go build tag for "experimental" or similar.

We could do this. A simpler approach would be to make that known in the godoc of the functions introduced. Thoughts?

decentralgabe:
agree with that

Issue #283, opened 2023-02-15
[Idea] DID Document Builder


andorsk:
**Is your feature request related to a problem? Please describe.**
Similar to how there is a VerifiableCredential builder: https://github.com/TBD54566975/ssi-sdk/blob/main/credential/builder.go#L24, I think it would be helpful to also have a DIDDocument builder. 

**Describe the solution you'd like**
Similar implementation to the VerifiableCredentialBuilder, but for DIDDocuments.

**Describe alternatives you've considered**
The alternative is to just expect to add methods via the struct. 

**Getting it done**
I'm considering building this out. If I do want to do this, I can contribute a PR to do this.

decentralgabe:
Done in #284

PR #284, opened 2023-02-15
Adding DID Document Builder

andorsk:
In reference to https://github.com/TBD54566975/ssi-sdk/issues/283


andorsk:
@decentralgabe can you take a look and merge if this is ok?

did/builder.go
```diff
@@ -0,0 +1,157 @@
...
+func (builder *DIDDocumentBuilder) AddContext(context interface{}) error {
+	if builder.IsEmpty() {
+		return errors.New(BuilderEmptyError)
+	}
+	res, err := util.InterfaceToStrings(context)
+	if err != nil {
+		return errors.Wrap(err, "malformed context")
+	}
+	uniqueContexts := util.MergeUniqueValues(builder.contexts, res)
+	builder.contexts = uniqueContexts
```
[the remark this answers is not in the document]
andorsk: great question. I used https://github.com/TBD54566975/ssi-sdk/blob/main/credential/builder.go as a reference. The reason it was suggested was the following: 

On : https://github.com/TBD54566975/ssi-sdk/blob/main/credential/builder.go#L25 	
// contexts and types are kept to avoid having cast to/from interface{} values

I made an assumption here since the context is also the same data type, this is a reasonable implementation to rely upon for did document builder.

did/builder_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"testing"
+
+	"github.com/TBD54566975/ssi-sdk/crypto"
+	"github.com/stretchr/testify/assert"
+)
+
+// Exercise all builder methods
+func TestDIDDocumentBuilder(t *testing.T) {
+	// https://www.w3.org/TR/did-core/#example-did-document-with-1-verification-method-type
+	var exampleAuthenticationEntry = map[string]string{
```
decentralgabe, line 13: please switch these var declarations to use `:=` syntax

decentralgabe: or add a single `var ( ) ` block

did/builder_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// https://www.w3.org/TR/did-core/#example-key-agreement-property-containing-two-verification-methods
+	var exampleKeyAgreementEntry = map[string]string{
+		"id":                 "did:example:123#zC9ByQ8aJs8vrNXyDhPHHNNMSHPcaSgNpjjsBYpMMjsTdS",
+		"type":               "X25519KeyAgreementKey2019",
+		"controller":         "did:example:123",
+		"publicKeyMultibase": "z9hFgmPVfmBZwRvFEyniQDBkz9LmV7gDEqytWyGZLmDXE",
+	}
+
+	builderBad := DIDDocumentBuilder{}
```
decentralgabe, line 61:
```suggestion
	var builderBad DIDDocumentBuilder
```

decentralgabe (approved):
supportive of this!

did/builder_test.go
andorsk: nice callout!

did/builder_test.go
andorsk: sounds good.

andorsk:
Thanks for the review @decentralgabe . addressed the comments. Realized you might be busy. No rush here. @nitro-neal lmk if your concerns are addressed as well. 

Thanks guys!

andorsk:
bonk. @decentralgabe or @nitro-neal can you get back and possible merge if you're good with this?

did/builder_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+import (
+	"testing"
+
+	"github.com/TBD54566975/ssi-sdk/crypto"
+	"github.com/stretchr/testify/assert"
+)
+
+// Exercise all builder methods
+func TestDIDDocumentBuilder(t *testing.T) {
```
andresuribe87, line 11: Would it make sense to split this test into multiple smaller ones, that exercise a single behavior?

andresuribe87 (commented):
I understand that this is following the same pattern that existed, but it is strange to me that every setter method returns an error, which can only be the BuilderEmptyError. I think it would be cleaner for only `Build` to return an error. The rest should simply accumulate changes.

2. The DIDDocument field is can be easily modified by clients, which

did/builder.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/TBD54566975/ssi-sdk/util"
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+)
+
+// contexts and types are kept to avoid having cast to/from interface{} values
+type DIDDocumentBuilder struct {
+	contexts []string
+	types    []string
+	*DIDDocument
```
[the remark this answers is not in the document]
andorsk: @andresuribe87 I was following precedence from previous code implemented in the ssi-sdk: https://github.com/TBD54566975/ssi-sdk/blob/main/credential/builder.go#L24.

did/builder.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func (builder *DIDDocumentBuilder) IsEmpty() bool {
+	if builder == nil || builder.DIDDocument == nil {
+		return true
+	}
+	return reflect.DeepEqual(builder, &DIDDocumentBuilder{})
+}
+
+func (builder *DIDDocumentBuilder) AddContext(context interface{}) error {
```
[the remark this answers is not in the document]
andorsk: @andresuribe87 as long as this works, I'm fine moving to any, but note https://github.com/TBD54566975/ssi-sdk/blob/main/credential/builder.go#L68 uses the `interface{}` type.

did/builder.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// Builds the DID Document
+func (builder *DIDDocumentBuilder) Build() (*DIDDocument, error) {
+	if builder.IsEmpty() {
+		return nil, errors.New(BuilderEmptyError)
+	}
+
+	if err := builder.DIDDocument.IsValid(); err != nil {
+		return nil, errors.Wrap(err, "did doc not ready to be built")
```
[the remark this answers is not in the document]
andorsk: sgtm. That works IMO.

did/builder.go
```diff
@@ -0,0 +1,158 @@
...
+			ID:      uuid.NewString(),
+			Context: contexts,
+		},
+	}
+}
+
+// Builds the DID Document
+func (builder *DIDDocumentBuilder) Build() (*DIDDocument, error) {
+	if builder.IsEmpty() {
+		return nil, errors.New(BuilderEmptyError)
```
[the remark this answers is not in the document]
andorsk: @andresuribe87 understand. Following existing implementation here https://github.com/TBD54566975/ssi-sdk/blob/main/credential/builder.go#L70 to stay consistent. If we want to revisit how errors are handle in the builders I'm OK with that, but I think they should be consistent across the VC builder and the DIDDocument builder.

did/builder_test.go
andorsk: Reasonable suggestion if we want to update that. I can put a note in for a future PR to move this into smaller tests.

andresuribe87: Why not do it in this PR?

From a maintainability perspective, it's easier to have smaller, self contained tests. 

FWIW, I want to add coding guidelines into the contribution doc. See https://github.com/TBD54566975/ssi-sdk/issues/290

did/builder.go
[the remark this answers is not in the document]
andresuribe87: Either works, filed https://github.com/TBD54566975/ssi-sdk/issues/289

did/builder_test.go
andorsk: @andresuribe87 again, this was all using the reference of the VC Builder code, which does tests in a similar way. https://github.com/TBD54566975/ssi-sdk/blob/main/credential/builder_test.go#L254. I was being consistent with the existing codebase. I think if we want to break things out, we should also address the VC Builder as well.

andresuribe87 (approved):

decentralgabe:
thanks @andorsk !

andorsk:
Thanks @andresuribe87 @nitro-neal @decentralgabe for the review!

PR #294, opened 2023-02-27
Uptake latest go

### LINKED ISSUE #291: Uptake go1.20+
This also involves migrating from `crypto/elliptic` to `crypto/ecdh`, which is safer and more efficient.

decentralgabe:
Fixes #291

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.19.6
+          go-version: 1.20.1
 
       - name: Install Mage
         run: go install github.com/magefile/mage
@@ -38,7 +38,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.19.6
+          go-version: 1.20.1
 
       - name: Install Mage
         run: go install github.com/magefile/mage
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -16,7 +16,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v3
         with:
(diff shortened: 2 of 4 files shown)


crypto/bbs.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"crypto/rand"
+	"crypto/sha256"
+
+	bbs "github.com/hyperledger/aries-framework-go/pkg/crypto/primitive/bbs12381g2pub"
+)
+
+// GenerateBBSKeyPair https://w3c-ccg.github.io/ldp-bbs2020
+func GenerateBBSKeyPair() (*bbs.PublicKey, *bbs.PrivateKey, error) {
```
andresuribe87, line 11: Did you mean to have this be part of this PR?

andresuribe87 (commented):
Are you intending to replace the `crypto/elliptic` occurrences in a different PR?

crypto/bbs.go
decentralgabe: nope

decentralgabe:
@andresuribe87 the update is for `The new [crypto/ecdh](https://go.dev/pkg/crypto/ecdh) package provides explicit support for Elliptic Curve Diffie-Hellman key exchanges over NIST curves and Curve25519.` the code we have doesn't use ECDH

andresuribe87:
@decentralgabe I was thinking of the crypto/keys.go file as below https://github.com/TBD54566975/ssi-sdk/blob/d48966db0f15c5cb32efcc4233d2eb5368adb1ee/crypto/keys.go#L93 ?

decentralgabe:
> https://go.dev/pkg/crypto/ecdh

Since we're not using ECDH (just ECDSA for now) there aren't any code changes to make. When/if we uptake ECDH we'll make use of this package.

andresuribe87 (approved):

PR #304, opened 2023-03-15
Add models for Credential Issuer Metadata

andresuribe87:
This is based in the spec section https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-issuer-metadata

I believe the SDK is the best home for this. My expectation is that other folks will need to support this kind of models, and providing a library to do so easily will be beneficial to the community. 

I thought about separating it into it's own module, but given that we're packaging all functionality into the SDK, I believe this is the best place for this to live.

### DIFF
--- a/oidc/issuance/metadata.go
+++ b/oidc/issuance/metadata.go
@@ -52,7 +52,7 @@ const (
 type CredentialSupported struct {
 	Format Format `json:"format" validate:"required"`
 
-	Id *string `json:"id,omitempty"`
+	ID *string `json:"id,omitempty"`
 
 	CryptographicBindingMethodsSupported []CryptographicBindingMethodSupported `json:"cryptographic_binding_methods_supported,omitempty"`
 
@@ -64,8 +64,8 @@ type CredentialSupported struct {
 	*JWTVCJSONCredentialMetadata
 }
 
-// DIDBindingMethods returns a list of all the did methods supported from the list of CryptographicBindingMethodsSupported.
-func (s CredentialSupported) DIDBindingMethods() []did.Method {
+// BindingDIDMethods returns a list of all the did methods supported from the list of CryptographicBindingMethodsSupported.
+func (s CredentialSupported) BindingDIDMethods() []did.Method {
 	methods := make([]did.Method, 0, len(s.CryptographicBindingMethodsSupported))
 	for _, bm := range s.CryptographicBindingMethodsSupported {
 		method, ok := bm.DIDBinding()
@@ -98,7 +98,7 @@ type IssuerMetadata struct {
 	// Must use the `https` scheme.
 	BatchCredentialEndpoint *util.URL `json:"batch_credential_endpoint,omitempty"`
 
-	CredentialsSupported []CredentialSupported `json:"credentials_supported,omitempty"`
+	CredentialsSupported CredentialSupported `json:"credentials_supported,omitempty"`
 
 	Display []displayJSON `json:"display,omitempty"`
 }
@@ -158,8 +158,7 @@ func (c Claim) MarshalJSON() ([]byte, error) {
 
 func (c *Claim) UnmarshalJSON(data []byte) error {
 	var cJSON claimJSON
-	err := json.Unmarshal(data, &cJSON)
-	if err != nil {
+	if err := json.Unmarshal(data, &cJSON); err != nil {
 		return errors.Wrap(err, "unmarshalling")
 	}
 	c.Mandatory = cJSON.Mandatory
--- a/oidc/issuance/metadata.go
+++ b/oidc/issuance/metadata.go
@@ -44,9 +44,9 @@ type CredentialDisplay struct {
 type Format string
 
 const (
-	JwtVcJSON   Format = "jwt_vc_json"
-	JwtVcJSONLd Format = "jwt_vc_json-ld"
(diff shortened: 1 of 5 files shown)


oidc/issuance/metadata_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	_ "embed"
+	"testing"
+
+	"github.com/goccy/go-json"
+	"github.com/stretchr/testify/require"
+)
+
+//go:embed example_issuer_metadata.json
```
nitro-neal, line 11: cool!

nitro-neal (approved):
👍

oidc/issuance/metadata.go
```diff
@@ -0,0 +1,167 @@
...
+type CryptographicBindingMethodSupported string
+
+const (
+	JWKFormat CryptographicBindingMethodSupported = "jwk"
+	COSEKey                                       = "cose_key"
+	AllDIDs                                       = "did"
+)
+
+type Logo struct {
+	Url     *util.URL `json:"url,omitempty"`
```
decentralgabe, line 19:
```suggestion
	URL     *util.URL `json:"url,omitempty"`
```

oidc/issuance/metadata.go
```diff
@@ -0,0 +1,167 @@
...
+	Description     *string `json:"description,omitempty"`
+	BackgroundColor *string `json:"background_color,omitempty"`
+	TextColor       *string `json:"text_color,omitempty"`
+}
+
+type Format string
+
+const (
+	JwtVcJson   Format = "jwt_vc_json"
+	JwtVcJsonLd        = "jwt_vc_json-ld"
```
decentralgabe, line 36:
```suggestion
	JWTVCJSONLD        = "jwt_vc_json-ld"
```

oidc/issuance/metadata.go
```diff
@@ -0,0 +1,167 @@
...
+	LdpVc              = "ldp_vc"
```
decentralgabe, line 37:
```suggestion
	LDPVC              = "ldp_vc"
```

oidc/issuance/metadata.go
```diff
@@ -0,0 +1,167 @@
...
+)
+
+type CredentialSupported struct {
+	Format Format `json:"format" validate:"required"`
+
+	Id *string `json:"id,omitempty"`
```
decentralgabe, line 43:
```suggestion
	ID *string `json:"id,omitempty"`
```
 
is a string ptr needed over a string?

oidc/issuance/metadata.go
```diff
@@ -0,0 +1,167 @@
...
+
+	return json.Marshal(jsonStruct)
+}
+
+func (c *Claim) UnmarshalJSON(data []byte) error {
+	var cJSON claimJSON
+	err := json.Unmarshal(data, &cJSON)
+	if err != nil {
+		return errors.Wrap(err, "unmarshalling")
+	}
```
decentralgabe, line 140:
```suggestion
	if err := json.Unmarshal(data, &cJSON); err != nil {
		return errors.Wrap(err, "unmarshalling")
	}
```

util/url.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/goccy/go-json"
+	"github.com/pkg/errors"
+)
+
+// URL is a wrapper struct of url.URL in order to unmarshal(marshal) URLs from(to) strings.
+type URL struct {
+	url.URL
+}
+
+func (u URL) MarshalJSON() ([]byte, error) {
```
decentralgabe, line 15: I believe you should be able to just call this Marshal and it will implement the json marshal interface

util/url.go
```diff
@@ -0,0 +1,31 @@
...
+	return json.Marshal(u.URL.String())
+}
+
+func (u *URL) UnmarshalJSON(data []byte) error {
+	var dataStr string
+	err := json.Unmarshal(data, &dataStr)
+	if err != nil {
+		return errors.Wrap(err, "unmarshalling")
+	}
```
decentralgabe, line 24:
```suggestion
	if err := json.Unmarshal(data, &dataStr); err != nil {
		return errors.Wrap(err, "unmarshalling")
	}
```

oidc/issuance/metadata.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/pkg/errors"
+	"golang.org/x/text/language"
+)
+
+type CryptographicBindingMethodSupported string
+
+const (
+	JWKFormat CryptographicBindingMethodSupported = "jwk"
+	COSEKey                                       = "cose_key"
+	AllDIDs                                       = "did"
```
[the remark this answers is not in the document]
andresuribe87: I didn't. This is part of what's specified in the spec. You made me realize that this approach was sub-par. So I added more documentation, and a new method to get all the Binding DID Methods from the `CredentialSupported` struct

oidc/issuance/metadata.go
decentralgabe: yeah it sucks

util/url.go
decentralgabe: looks like you had it right my bad http://choly.ca/post/go-json-marshalling/

decentralgabe (approved):

PR #307, opened 2023-03-17
Ion sdk tools

decentralgabe:
Add feature parity to the following repos:
- https://github.com/decentralized-identity/ion-sdk
- https://github.com/decentralized-identity/ion-tools


decentralgabe:
depends on https://github.com/TBD54566975/ssi-sdk/pull/308

did/ion/model.go
```diff
@@ -0,0 +1,99 @@
...
+type UpdateRequest struct {
+	Type        OperationType `json:"type,omitempty"`
+	DIDSuffix   string        `json:"didSuffix,omitempty"`
+	RevealValue string        `json:"revealValue,omitempty"`
+	Delta       Delta         `json:"delta,omitempty"`
+	SignedData  string        `json:"signedData,omitempty"`
+}
+
+type Delta struct {
+	Patches          []any  `json:"patches,omitempty"`
```
andresuribe87, line 82: Can you clarify in a comment what the possible types are? 

I suggest unexporting this field, and having typed methods that you can call to add a patch. E.g. `func (d *Delta) AddReplaceAction(r ReplaceAction) *Delta`

did/ion/request.go
```diff
@@ -0,0 +1,302 @@
...
+	ServicesToAdd        []Service
+	ServiceIDsToRemove   []string
+	PublicKeysToAdd      []PublicKey
+	PublicKeyIDsToRemove []string
+}
+
+func (s StateChange) IsValid() error {
+	// check if services are valid
+	// build index of services to make sure IDs are unique
+	services := make(map[string]Service)
```
andresuribe87, line 157: No need to map to actual services. Also, hint at the right size. 
```suggestion
	services := make(map[string]struct{}, len(s.ServicesToAdd))
```

did/ion/request.go
```diff
@@ -0,0 +1,302 @@
...
+		if len(service.Type) > maxServiceTypeLength {
+			return fmt.Errorf("service<%s> type %s is too long", service.ID, service.Type)
+		}
+
+		services[service.ID] = service
+	}
+
+	// check if public keys are valid
+	// build index of public keys to add
+	publicKeys := make(map[string]PublicKey)
```
andresuribe87, line 177: Same as above.
```suggestion
	publicKeys := make(map[string]struct{}, len(s.PublicKeysToAdd))
```

did/ion/request.go
```diff
@@ -0,0 +1,302 @@
...
+	for _, publicKey := range s.PublicKeysToAdd {
+		if _, ok := publicKeys[publicKey.ID]; ok {
+			return fmt.Errorf("public key<%s> is duplicated", publicKey.ID)
```
andresuribe87, line 180: Here and elsewhere. If you want stack traces, use `errors.Errorf`. See https://stackoverflow.com/questions/61933650/whats-the-difference-between-errors-wrapf-errors-errorf-and-fmt-errorf for more details. (make sure it's **not** the native `"errors"` library).

did/ion/request.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// NewUpdateRequest creates a new update request https://identity.foundation/sidetree/spec/#update
+func NewUpdateRequest(didSuffix string, updateKey, nextUpdateKey crypto.PublicKeyJWK, signer BTCSignerVerifier, stateChange StateChange) (*UpdateRequest, error) {
+	if err := stateChange.IsValid(); err != nil {
+		return nil, err
+	}
+
+	// construct update patches
+	var patches []any
```
andresuribe87, line 221: Hint the container capacity
```suggestion
	patches := make([]any, 0, len(stateChange.ServicesToAdd) + ...)
```

did/ion/request.go
```diff
@@ -0,0 +1,302 @@
...
+	deltaHash, err := HashEncode(deltaCanonical)
+	if err != nil {
+		return nil, err
+	}
+
+	// prepare signed data
+	toBeSigned := struct {
+		UpdateKey crypto.PublicKeyJWK `json:"updateKey"`
+		DeltaHash string              `json:"deltaHash"`
+	}{
```
andresuribe87, line 287: Curious why an anonymous struct here vs defining it in `model.go` ?

did/ion/request.go
```diff
@@ -0,0 +1,302 @@
...
+			return fmt.Errorf("public key<%s> id is too long", publicKeyID)
+		}
+	}
+	return nil
+}
+
+// NewUpdateRequest creates a new update request https://identity.foundation/sidetree/spec/#update
+func NewUpdateRequest(didSuffix string, updateKey, nextUpdateKey crypto.PublicKeyJWK, signer BTCSignerVerifier, stateChange StateChange) (*UpdateRequest, error) {
+	if err := stateChange.IsValid(); err != nil {
+		return nil, err
```
andresuribe87, line 217: Here and elsewhere in this func. I recommend chaining errors with either `Wrap` or using `:` and `Errorf`. See https://github.com/uber-go/guide/blob/master/style.md#error-wrapping for more details.

did/ion/crypto.go
```diff
@@ -0,0 +1,222 @@
...
+
+	return reveal, commitment, nil
+}
+
+type BTCSignerVerifier struct {
+	publicKey  *btcec.PublicKey
+	privateKey *btcec.PrivateKey
+}
+
+// NewBTCSignerVerifier creates a new signer/verifier for signatures suited for the BTC blockchain
```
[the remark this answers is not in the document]
decentralgabe: not a specific interface, just a utility to wrap the key

did/ion/crypto_test.go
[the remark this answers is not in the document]
decentralgabe: they are all tested through the test vectors, though not directly

did/ion/did.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+type InitialState struct {
+	SuffixData SuffixData `json:"suffixData,omitempty"`
+	Delta      Delta      `json:"delta,omitempty"`
+}
+
+func CreateLongFormDID(recoveryKey, updateKey crypto.PublicKeyJWK, document Document) (string, error) {
```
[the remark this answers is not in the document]
decentralgabe: added

did/ion/did_test.go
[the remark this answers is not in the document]
decentralgabe: adding

did/ion/did_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+// https://github.com/decentralized-identity/ion-sdk/blob/main/tests/IonDid.spec.ts#L18
+func TestCreateLongFormDID(t *testing.T) {
+	recoveryKeyJSON, err := getTestData("jwkes256k1public.json")
+	assert.NoError(t, err)
+	var recoveryKey crypto.PublicKeyJWK
+	err = json.Unmarshal([]byte(recoveryKeyJSON), &recoveryKey)
+	assert.NoError(t, err)
```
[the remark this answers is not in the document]
decentralgabe: will simplify this

did/ion/model.go
decentralgabe: I had something like this at first but the marshalling was getting some errors..let me see

did/ion/request.go
decentralgabe: ah was all anonymous first, forgot to update this

did/ion/request.go
decentralgabe: updated

did/ion/model.go
decentralgabe: added helper methods. the field cannot be unexported otherwise itll be dropped during serialization

did/ion/did.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/TBD54566975/ssi-sdk/crypto"
+	"github.com/TBD54566975/ssi-sdk/did"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+type InitialState struct {
+	SuffixData SuffixData `json:"suffixData,omitempty"`
+	Delta      Delta      `json:"delta,omitempty"`
+}
```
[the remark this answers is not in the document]
decentralgabe: no - it is needed for short form did too

did/ion/model.go
andresuribe87: It's the classic oneof problem with golang, which has no great support :( 

We could use the [protobuf approach](https://protobuf.dev/reference/go/go-generated/#oneof). It's a bit of additional work, but I think it's worth it. What are your thoughts?

andresuribe87 (approved):
Not blocking, but part of an important convo.

decentralgabe:
@andresuribe87 I agree protobufs would work. I think JSON schemas would be a simpler solution.

since the source of truth is going to be an ion node (which will reject an invalid request) I think we are OK for now, but should think about it as an improvement

PR #329, opened 2023-03-28
add context to resolver

decentralgabe:
unifies resolution interface between the sdk and service

### DIFF
--- a/did/key_fuzz_test.go
+++ b/did/key_fuzz_test.go
@@ -1,6 +1,7 @@
 package did
 
 import (
+	"context"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -46,7 +47,7 @@ func FuzzCreateAndResolve(f *testing.F) {
 		didKey, err := CreateDIDKey(kt, pubKey)
 		assert.NoError(t, err)
 
-		doc, err := resolver.Resolve(didKey.String())
+		doc, err := resolver.Resolve(context.TODO(), didKey.String())
 		if err != nil {
 			t.Skip()
 		}
--- a/did/resolver.go
+++ b/did/resolver.go
@@ -1,6 +1,7 @@
 package did
 
 import (
+	"context"
 	"fmt"
 	"strings"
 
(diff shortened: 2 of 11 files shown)


decentralgabe:
will uptake https://github.com/TBD54566975/ssi-sdk/pull/329 before merge

did/resolver.go
```diff
@@ -1,20 +1,21 @@
...
 	"strings"
 
 	"github.com/goccy/go-json"
 	"github.com/pkg/errors"
 )
 
-// ResolutionOptions https://www.w3.org/TR/did-spec-registries/#did-resolution-options
-type ResolutionOptions any
+// ResolutionOption https://www.w3.org/TR/did-spec-registries/#did-resolution-options
+type ResolutionOption any
```
andresuribe87, line 13: Whenever I see `any`, it's typically a symptom of an API that hasn't been fleshed out yet. In this case, it's unclear the shape of what `ResolutionOption` is acceptable by the functions that use it. 

Instead, I would define an interface. Mind creating an issue to address?

Separately, This doesn't seem to be used anywhere.

andresuribe87 (commented):
While you're here, couple more suggestions I have:
* Rename the `did.Resolution` interface to `did.Resolver`. This makes it clear that anyone that implementers will need to implement `Resolve`. 
* Rename the `did.Resolver` struct to `did.MultiMethodResolver`.

did/key_fuzz_test.go
```diff
@@ -46,7 +47,7 @@ func FuzzCreateAndResolve(f *testing.F) {
 		didKey, err := CreateDIDKey(kt, pubKey)
 		assert.NoError(t, err)
 
-		doc, err := resolver.Resolve(didKey.String())
+		doc, err := resolver.Resolve(context.TODO(), didKey.String())
```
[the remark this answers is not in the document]
decentralgabe: historically I had used TODO for tests to signify it's not a real context, but happy to switch.

did/resolver.go
decentralgabe: yeah not used anywhere yet...will create an issue

decentralgabe: https://github.com/TBD54566975/ssi-sdk/issues/331

decentralgabe:
addressed all comments - switched all TODO to Background

PR #332, opened 2023-03-28
multi method resolver

decentralgabe:
update the resolver interface to support multiple methods per resolver as an option.
enforced interface compliance for resolvers in the sdk.

### DIFF
--- a/did/resolver.go
+++ b/did/resolver.go
@@ -16,7 +16,7 @@ type ResolutionOption any
 type Resolver interface {
 	// Resolve Attempts to resolve a DID for a given method
 	Resolve(ctx context.Context, did string, opts ...ResolutionOption) (*ResolutionResult, error)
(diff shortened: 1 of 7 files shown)


did/resolver.go
```diff
@@ -16,8 +16,8 @@ type ResolutionOption any
 type Resolver interface {
 	// Resolve Attempts to resolve a DID for a given method
 	Resolve(ctx context.Context, did string, opts ...ResolutionOption) (*ResolutionResult, error)
-	// Method provides the method for the given resolution implementation
-	Method() Method
+	// Methods provides the method(s) for the given resolution implementation
```
andresuribe87, line 19:
```suggestion
	// Methods returns all methods that can be resolved by this resolver.
```

andresuribe87 (approved):
Tiny suggestion

PR #343, opened 2023-04-10
Add JWT header extraction; fix errs

decentralgabe:
- support extracting headers from JWTs, useful for getting `kid` and other info needed in verification
- for a logging err, check if err is nil, which was resulting in returning a nil value when an error was expected

### DIFF
--- a/credential/signing/jwt.go
+++ b/credential/signing/jwt.go
@@ -8,6 +8,7 @@ import (
 	"github.com/TBD54566975/ssi-sdk/crypto"
 	"github.com/goccy/go-json"
 	"github.com/google/uuid"
+	"github.com/lestrrat-go/jwx/v2/jws"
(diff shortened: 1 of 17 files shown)


credential/signing/jwt.go
```diff
@@ -277,5 +290,16 @@ func ParseVerifiablePresentationFromJWT(token string) (jwt.Token, *credential.Ve
...
-	return parsed, &pres, nil
+	return headers, parsed, &pres, nil
+}
+
+func getJWTHeaders(token []byte) (jws.Headers, error) {
+	msg, err := jws.Parse(token)
+	if err != nil {
+		return nil, err
+	}
+	if len(msg.Signatures()) != 1 {
```
nitro-neal, line 301: are there other checks that we can do to make sure these are valid headers?

nitro-neal (approved):

PR #353, opened 2023-04-14
Add signature checks for Verifiable Presentations

### LINKED ISSUE #71: Presentation Exchange: validate claim signatures in presentation submissions
Presentation submissions may contain claims, and may contain signatures over those claims. The SDK should be updated to support validating the signatures on claims submitted as a part of a Presentation Submission.

decentralgabe:
Fixes #71
- make all signer/verifier methods generic
- shuffle packages to remove cycles
- add signature checks for VPs
- move in method to resolve a key given a DID and KID
- disabled improperly implemented limited disclosure, opened https://github.com/TBD54566975/ssi-sdk/issues/354
- had to refactor the test cases to accomodate real signed VCs / verifiers, updated the test wallet accordingly


credential/signature.go
```diff
@@ -0,0 +1,106 @@
...
+	"reflect"
+
+	"github.com/TBD54566975/ssi-sdk/crypto"
+	"github.com/TBD54566975/ssi-sdk/did"
+	"github.com/pkg/errors"
+)
+
+// VerifyCredentialSignature verifies the signature of a credential of any type
+// TODO(gabe) support other types of credentials https://github.com/TBD54566975/ssi-sdk/issues/352
+func VerifyCredentialSignature(ctx context.Context, genericCred any, resolver did.Resolver) (bool, error) {
```
decentralgabe, line 16: important stuff here

credential/exchange/verification.go
```diff
@@ -30,7 +34,12 @@ func VerifyPresentationSubmission(verifier crypto.JWTVerifier, et EmbedTarget, d
 	}
 	switch et {
 	case JWTVPTarget:
-		_, _, vp, err := signing.VerifyVerifiablePresentationJWT(verifier, string(submission))
+		jwtVerifier, ok := verifier.(crypto.JWTVerifier)
+		if !ok {
+			return fmt.Errorf("verifier<%T> is not a JWT verifier", verifier)
+		}
+		// verify the VP, which in turn verifies all credentials in it
+		_, _, vp, err := credential.VerifyVerifiablePresentationJWT(context.Background(), jwtVerifier, resolver, string(submission))
```
nitro-neal, line 42: eventually context.background should be top level

credential/jwt.go
```diff
@@ -245,23 +248,60 @@ func SignVerifiablePresentationJWT(signer crypto.JWTSigner, parameters JWTVVPPar
...
-// https://www.w3.org/TR/vc-data-model/#jwt-decoding
-// If there are any issues during decoding, an error is returned. As a result, a successfully
-// decoded VerifiablePresentation object is returned.
-func VerifyVerifiablePresentationJWT(verifier crypto.JWTVerifier, token string) (jws.Headers, jwt.Token, *credential.VerifiablePresentation, error) {
+// VerifyVerifiablePresentationJWT verifies the signature validity on the token. Then, the JWT is decoded according
+// to the specification: https://www.w3.org/TR/vc-data-model/#jwt-decoding
+// After decoding the signature of each credential in the presentation is verified. If there are any issues during
+// decoding or signature validation, an error is returned. As a result, a successfully decoded VerifiablePresentation
+// object is returned.
+func VerifyVerifiablePresentationJWT(ctx context.Context, verifier crypto.JWTVerifier, resolver did.Resolver, token string) (jws.Headers, jwt.Token, *VerifiablePresentation, error) {
```
nitro-neal, line 256: :thumbsup:

credential/signature.go
```diff
@@ -0,0 +1,110 @@
...
+	if genericCred == nil {
+		return false, errors.New("credential cannot be empty")
+	}
+	if resolver == nil {
```
nitro-neal, line 20: if the resolver was empty could there be a default one we load?

credential/signature.go
```diff
@@ -0,0 +1,110 @@
...
+		return false, errors.New("resolver cannot be empty")
+	}
+	switch genericCred.(type) {
+	case *VerifiableCredential:
+		return VerifyCredentialSignature(ctx, *genericCred.(*VerifiableCredential), resolver)
+	case VerifiableCredential:
+		cred := genericCred.(VerifiableCredential)
```
nitro-neal, line 27: no ok checking here?
cred, ok := genericCred.(VerifiableCredential)
I guess we know the type so not needed

credential/signature.go
```diff
@@ -0,0 +1,110 @@
...
+
+		// first try as a VC object
+		var cred VerifiableCredential
+		if err = json.Unmarshal(credMapBytes, &cred); err == nil && !cred.IsEmpty() {
+			return VerifyCredentialSignature(ctx, cred, resolver)
+		}
+
+		// if that fails, try as a JWT
+		if _, _, _, err := VCJWTJSONToVC(credMapBytes); err == nil {
+			return false, errors.New("JWT credentials must include a signature to be verified")
```
nitro-neal, line 63: Is this the right error message here?

nitro-neal: It could just be an invalid map[string]any

credential/util.go
```diff
@@ -0,0 +1,128 @@
...
+	"reflect"
+
+	"github.com/goccy/go-json"
+	"github.com/lestrrat-go/jwx/v2/jws"
+	"github.com/lestrrat-go/jwx/v2/jwt"
+	"github.com/pkg/errors"
+)
+
+// ToCredential turn a generic cred into its known object model
+func ToCredential(genericCred any) (jws.Headers, jwt.Token, *VerifiableCredential, error) {
```
nitro-neal, line 14: A lot of code reuse, would it be possible to Call this funciton in the signature.go VerifyCredentialSignature?

example/usecase/employer_university_flow/pkg/verifier.go
```diff
@@ -1,41 +1,27 @@
...
 
 // ValidateAccess is a very simple validation process against a Presentation Submission
 // It checks:
-// 1. That the VC is valid
-// 2. That the VC was issued by a trusted entity
-func ValidateAccess(verifier crypto.JWTVerifier, credBytes []byte) error {
-	_, _, vp, err := signing.VerifyVerifiablePresentationJWT(verifier, string(credBytes))
+// 1. That the VP is valid
+// 2. All VCs in the VP are valid
+// 2. That the VC was issued by a trusted entity (implied by the presentation, according to the Presentation Definition)
```
nitro-neal, line 16: 3.

nitro-neal (commented):
.

credential/signature.go
```diff
@@ -0,0 +1,110 @@
...
+func VerifyCredentialSignature(ctx context.Context, genericCred any, resolver did.Resolver) (bool, error) {
+	if genericCred == nil {
+		return false, errors.New("credential cannot be empty")
+	}
+	if resolver == nil {
+		return false, errors.New("resolver cannot be empty")
+	}
+	switch genericCred.(type) {
+	case *VerifiableCredential:
+		return VerifyCredentialSignature(ctx, *genericCred.(*VerifiableCredential), resolver)
```
nitro-neal, line 25: Are we sure we won't run into an infinite loop problem here or at the other 
return VerifyCredentialSignature spots?

nitro-neal (approved):
Looks great! Just a few spots to look over and think about:
Biggest stuff is perhaps infinite loop / stack overflow problem if we are not careful with the return VerifyCredentialSignature

With the context, I wonder if we can extract it away and it will only be created once in the resolver or something

credential/exchange/verification.go
decentralgabe: I agree, I will make it top level here

credential/signature.go
decentralgabe: I'm not sure if we should, because that would imply supporting DID methods that the caller may not want to resolve

credential/signature.go
decentralgabe: yeah since the switch was hit we know the type, I can make this nicer by extracting the type above

decentralgabe: I'm not sure how to make it nicer without some crazy generic stuff

credential/signature.go
decentralgabe: sure...maybe not. I have each case covered in a test that should be "good enough"

example/usecase/employer_university_flow/pkg/verifier.go
decentralgabe: ty

credential/util.go
decentralgabe: good idea, I think it can be used for everything but the JWT case

Issue #146, opened 2022-07-15
Support for Post-Quantum Signature Types


decentralgabe:
https://datatracker.ietf.org/doc/draft-prorock-cose-post-quantum-signatures/

decentralgabe:
Go impl of Kyber! https://github.com/cloudflare/circl/tree/main/kem

michaelneale:
the "kyber crystal" name really does make it hard to take seriously ;)

decentralgabe:
Opened https://github.com/lestrrat-go/jwx/issues/912 which is needed for support for our jw* code

decentralgabe:
There is an incompatibility issue between the go and js dilithium suites which must be resolved before test vectors can be implemented https://github.com/cloudflare/circl/issues/435

PR #364, opened 2023-04-28
PQC Support for Dilithium

decentralgabe:
Fist attempt at addressing #146


decentralgabe:
JWS signing with dilithium is not yet supported tracking in https://github.com/lestrrat-go/jwx/issues/912

crypto/dilithium.go
```diff
@@ -0,0 +1,175 @@
...
+)
+
+func GenerateDilithiumKeyPair(m DilithiumMode) (dilithium.PublicKey, dilithium.PrivateKey, error) {
+	mode := dilithium.ModeByName(m.String())
+	if mode == nil {
+		return nil, nil, fmt.Errorf("unsupported dilithium mode: %s", m)
+	}
+	pk, sk, err := mode.GenerateKey(nil)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "could not generate key for dilithium")
```
andresuribe87, line 31:
```suggestion
		return nil, nil, errors.Wrap(err, "generating key for dilithium")
```

crypto/dilithium.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	return pk, sk, nil
+}
+
+type DilithiumSigner struct {
+	kid  string
+	mode dilithium.Mode
+	dilithium.PublicKey
+	dilithium.PrivateKey
```
andresuribe87, line 40: Are you intending for the `DilithiumSigner` to inherit the behaviors of `dilithium.PublicKey` and `dilithium.PrivateKey`?

If not, I think it improves readability to have these fields explicitly named, rather than doing a struct embedding.

crypto/dilithium.go
```diff
@@ -0,0 +1,175 @@
...
+	dilithium.PublicKey
+}
+
+func NewDilithiumVerifier(kid string, m DilithiumMode, pubKey dilithium.PublicKey) (*DilithiumVerifier, error) {
+	mode := dilithium.ModeByName(m.String())
+	if mode == nil {
+		return nil, fmt.Errorf("unsupported dilithium mode: %s", m)
+	}
+	// verify the mode matches the pk, this will panic if they don't match
+	_ = mode.PublicKeyFromBytes(pubKey.Bytes())
```
andresuribe87, line 79: Panic here seems dangerous. Would it make sense to return an error instead?

crypto/dilithium.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (s *DilithiumSigner) Sign(message []byte) []byte {
+	return s.mode.Sign(s.PrivateKey, message)
+}
+
+type DilithiumVerifier struct {
+	KID  string
+	mode dilithium.Mode
+	dilithium.PublicKey
```
andresuribe87, line 70: Same comment as above about making this explicit.

crypto/dilithium.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (s *DilithiumSigner) GetKeyID() string {
+	return s.kid
+}
+
+func (s *DilithiumSigner) Sign(message []byte) []byte {
+	return s.mode.Sign(s.PrivateKey, message)
+}
+
+type DilithiumVerifier struct {
```
andresuribe87, line 67: Should this type implement `cryptosuite.Verifier`?

crypto/dilithium.go
```diff
@@ -0,0 +1,175 @@
...
+		return nil, nil, fmt.Errorf("unsupported dilithium mode: %s", m)
+	}
+	pk, sk, err := mode.GenerateKey(nil)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "could not generate key for dilithium")
+	}
+	return pk, sk, nil
+}
+
+type DilithiumSigner struct {
```
andresuribe87, line 36: Should this type implement `cryptosuite.Signer`?

crypto/dilithium.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	return "", errors.New("unsupported dilithium mode")
+}
+
+func getMode2PrivateKey(privKey dilithium.PrivateKey) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = errors.New("not mode 2")
+		}
+	}()
```
andresuribe87, line 109: Here and elsewhere - I would recommend doing the check that's causing the panic. It would avoid recovering from a potentially really bad situation (like OOM for example).

crypto/dilithium_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+		pk, sk, err = GenerateDilithiumKeyPair(Dilithium3)
+		assert.NoError(tt, err)
+		assert.NotNil(tt, pk)
+		assert.NotNil(tt, sk)
+
+		pk, sk, err = GenerateDilithiumKeyPair(Dilithium5)
+		assert.NoError(tt, err)
+		assert.NotNil(tt, pk)
+		assert.NotNil(tt, sk)
```
andresuribe87, line 24: Consider using a test table

crypto/dilithium_test.go
```diff
@@ -0,0 +1,141 @@
+package crypto
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDilithiumSignVerify(t *testing.T) {
+	t.Run("Generate all possible keys per mode", func(tt *testing.T) {
```
andresuribe87, line 10: It's a good practice to name tests for each **behavior** that you want to test. See [this](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#naming-your-tests) for a more detailed explanation with examples.

As it stands, the intent of many of the tests is ambiguous and puts the onus of the reader.

crypto/jwx/jws_dilithium.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+func init() {
+	jws.RegisterSigner(DilithiumMode2Alg, jws.SignerFactoryFn(NewDilithiumMode2Signer))
+	jws.RegisterVerifier(DilithiumMode2Alg, jws.VerifierFactoryFn(NewDilithiumMode2Verifier))
+	jws.RegisterSigner(DilithiumMode3Alg, jws.SignerFactoryFn(NewDilithiumMode3Signer))
+	jws.RegisterVerifier(DilithiumMode3Alg, jws.VerifierFactoryFn(NewDilithiumMode3Verifier))
+	jws.RegisterSigner(DilithiumMode5Alg, jws.SignerFactoryFn(NewDilithiumMode5Signer))
+	jws.RegisterVerifier(DilithiumMode5Alg, jws.VerifierFactoryFn(NewDilithiumMode5Verifier))
+}
```
andresuribe87, line 18: Users of the SDK may not want to accept PQC yet. Thus, I recommend that this registration should be up to the user of the library, instead of an `init` function.

crypto/jwx/jws_dilithium.go
```diff
@@ -0,0 +1,86 @@
...
+	case dilithium.Mode3:
+		return DilithiumMode3Alg
+	case dilithium.Mode5:
+		return DilithiumMode5Alg
+	default:
+		return ""
+	}
+}
+
+func (s DilithiumSignerVerifier) Sign(payload []byte, keyif interface{}) ([]byte, error) {
```
andresuribe87, line 67:
```suggestion
func (s DilithiumSignerVerifier) Sign(payload []byte, key any) ([]byte, error) {
```

crypto/jwx/jws_dilithium.go
```diff
@@ -0,0 +1,86 @@
...
+	switch key := keyif.(type) {
+	case dilithium.PrivateKey:
+		return s.m.Sign(key, payload), nil
+	default:
+		return nil, fmt.Errorf(`invalid key type %T`, keyif)
+	}
+}
+
+func (s DilithiumSignerVerifier) Verify(payload []byte, signature []byte, keyif interface{}) error {
```
andresuribe87, line 76:
```suggestion
func (s DilithiumSignerVerifier) Verify(payload []byte, signature []byte, key any) error {
```

did/ion/testutil_test.go
```diff
@@ -13,8 +13,8 @@ var (
 	testData embed.FS
 )
 
-// RetrieveTestVectorAs retrieves a test vector from the testdata folder and unmarshals it into the given interface
-func RetrieveTestVectorAs(t *testing.T, fileName string, output interface{}) {
+// retrieveTestVectorAs retrieves a test vector from the testdata folder and unmarshals it into the given interface
+func retrieveTestVectorAs(t *testing.T, fileName string, output interface{}) {
```
andresuribe87, line 17: This function looks like it's repeated in another test. Could this be DRY'd up?

andresuribe87 (commented):
High level comments:
* I think we should add comments for all exported functions, as this is an SDK that we expect users to import. This will make it easier for devs to adopt it.
* There is a `crypto.DilithiumSigner`, a `crypto.DilithiumVerifier` and a `jwx.DilithiumSigner`. Is it possible to simplify this by removing the last one? If not, then it would make sense to add guidance to devs on which one to use under which scenarios.

Separately, reviewers would be grateful in future PRs if you can make them more review friendly :) . A couple of suggestions:
* Splitting the PR into smaller pieces (like refactoring vs adding features)
* Making sure that you use `git mv` to move files. This makes it easier to understand what changed vs what moved only.

crypto/dilithium.go
decentralgabe: no since it's not a part of any cryptosuite

crypto/dilithium.go
decentralgabe: behaviors...not exactly

if you see the sign method
```go
func (s *DilithiumSigner) Sign(message []byte) []byte {
	return s.mode.Sign(s.PrivateKey, message)
}
```
the behavior is defined on the `mode` not the key itself

crypto/dilithium.go
decentralgabe: no - same answer as above - not implementing a cryptosuite (now called a data integrity suite)

crypto/dilithium.go
decentralgabe: I'll add recovery logic

crypto/dilithium.go
decentralgabe: can you expand on how I'd do this, since the library throws a panic and does not return an error how would we avoid it?

decentralgabe: do you mean move

```go
	if len(data) != mode2.PrivateKeySize {
		panic("packed public key must be of mode2.PrivateKeySize bytes")
	}
```

from the library in here and then forget about panics?

decentralgabe: I will do this but generally I don't like intercepting what a method from a lib will do since it could change out from underneath us

crypto/dilithium_test.go
decentralgabe: good callout - will update

crypto/dilithium_test.go
decentralgabe: updated

crypto/jwx/jwk.go
```diff
@@ -51,22 +58,43 @@ func (k PrivateKeyJWK) ToPublicKeyJWK() PublicKeyJWK {
...
+func (k PrivateKeyJWK) toDilithiumPrivateKey() (gocrypto.PrivateKey, error) {
+	decodedPrivKey, err := base64.RawURLEncoding.DecodeString(k.D)
+	if err != nil {
+		return nil, err
+	}
+	switch k.Alg {
+	case DilithiumMode2Alg.String():
+		return dilithium.Mode2.PrivateKeyFromBytes(decodedPrivKey), nil
+	case DilithiumMode3Alg.String():
+		return dilithium.Mode3.PrivateKeyFromBytes(decodedPrivKey), nil
```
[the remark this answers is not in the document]
decentralgabe: will add them separately - the test vector only covers one mode

crypto/jwx/jwk.go
```diff
@@ -51,22 +58,43 @@ func (k PrivateKeyJWK) ToPublicKeyJWK() PublicKeyJWK {
...
 	}
 	// dereference the ptr
 	if reflect.ValueOf(goKey).Kind() == reflect.Ptr {
-		goKey = reflect.ValueOf(goKey).Elem().Interface().(crypto.PrivateKey)
+		goKey = reflect.ValueOf(goKey).Elem().Interface().(gocrypto.PrivateKey)
 	}
 	return goKey, nil
 }
 
+func (k PrivateKeyJWK) toDilithiumPrivateKey() (gocrypto.PrivateKey, error) {
```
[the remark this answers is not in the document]
decentralgabe: added

crypto/jwx/jwk.go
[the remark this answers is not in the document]
decentralgabe: the ones you noticed are for the test vectors in the spec - which only cover mode 5.

I have the other modes tested above in the test `TestJWKToPrivateKeyJWK`

crypto/jwx/jwk.go
```diff
@@ -81,22 +109,43 @@ type PublicKeyJWK struct {
...
 	// dereference the ptr
 	if reflect.ValueOf(goKey).Kind() == reflect.Ptr {
-		goKey = reflect.ValueOf(goKey).Elem().Interface().(crypto.PublicKey)
+		goKey = reflect.ValueOf(goKey).Elem().Interface().(gocrypto.PublicKey)
 	}
 	return goKey, nil
 }
 
+func (k PublicKeyJWK) toDilithiumPublicKey() (gocrypto.PublicKey, error) {
+	decodedPubKey, err := base64.RawURLEncoding.DecodeString(k.X)
```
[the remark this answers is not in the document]
decentralgabe: yes will add

crypto/jwx/jwk.go
```diff
@@ -180,16 +229,25 @@ func PublicKeyToPublicKeyJWK(key crypto.PublicKey) (*PublicKeyJWK, error) {
...
+		pubKey := dilithium.Mode5.PublicKeyFromBytes(k.Bytes())
+		return jwkFromDilithiumPublicKey(crypto.Dilithium5, pubKey)
 	default:
 		return nil, fmt.Errorf("unsupported public key type: %T", k)
 	}
 }
 
 // PrivateKeyToJWK converts a private key to a JWK
-func PrivateKeyToJWK(key crypto.PrivateKey) (jwk.Key, error) {
+func PrivateKeyToJWK(key gocrypto.PrivateKey) (jwk.Key, error) {
```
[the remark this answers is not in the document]
decentralgabe: https://github.com/TBD54566975/ssi-sdk/issues/365

crypto/jwx/jws_dilithium.go
decentralgabe: overhead is small. they may not want any key type, should they have to enable them individually?

this is an opinionated lib that provides broad support. my take is keep it as is.

go.mod
```diff
@@ -31,6 +31,7 @@ require (
 )
 
 require (
+	github.com/cloudflare/circl v1.3.2 // indirect
```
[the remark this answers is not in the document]
decentralgabe: done

did/ion/testutil_test.go
decentralgabe: not in an elegant way - no because `_test.go` files cannot be shared across packages. and we shouldn't bundle test code in the sdk

decentralgabe:
> * I think we should add comments for all exported functions, as this is an SDK that we expect users to import. This will make it easier for devs to adopt it.

Added 

> * There is a `crypto.DilithiumSigner`, a `crypto.DilithiumVerifier` and a `jwx.DilithiumSigner`. Is it possible to simplify this by removing the last one? If not, then it would make sense to add guidance to devs on which one to use under which scenarios.

The first two are all we'd use when not signing JW*. The JWX signer is needed to comply with the jwx lib we use.

> Separately, reviewers would be grateful in future PRs if you can make them more review friendly :) . A couple of suggestions:
> 
> * Splitting the PR into smaller pieces (like refactoring vs adding features)
> * Making sure that you use `git mv` to move files. This makes it easier to understand what changed vs what moved only.

Agree - will be more mindful of this going forward. Didn't know about `git mv`.

andresuribe87 (approved):
Thanks - small nits in the open comments.

PR #370, opened 2023-05-05
include dilithium; clean up key types and related methods

decentralgabe:


crypto/keys_test.go
```diff
@@ -70,4 +103,133 @@ func TestKeyToBytes(t *testing.T) {
...
+					assert.Nil(ttt, pk)
+					assert.Nil(ttt, sk)
+				} else {
+					assert.NoError(ttt, err)
+					assert.NotNil(ttt, pk)
+					assert.NotNil(ttt, sk)
+				}
+			})
+		}
+	})
```
andresuribe87, line 182: nit: minimize branching logic inside tests. In this case I suggest separating the error case behavior into a separate test. It would make things more readable.

PR #371, opened 2023-05-08
Refactor DID package

decentralgabe:
- move each did method into its own directory
- break out resolver code into new file
- make usage of shared utils a bit more clear

no new code - just refactoring


did/resolver/resolver.go
```diff
@@ -1,12 +1,15 @@
-package did
+package resolver
```
andresuribe87, line 1: To avoid stuttering, and seems like a better name for the package. 
```suggestion
package resolution
```

did/resolver/resolver_test.go
```diff
@@ -0,0 +1,44 @@
...
+	var metadata DocumentMetadata
+	assert.True(t, metadata.IsValid())
+
+	// bad
+	badMetadata := DocumentMetadata{
+		Created: "bad",
+		Updated: time.Now().UTC().Format(time.RFC3339),
+	}
+	assert.False(t, badMetadata.IsValid())
+}
```
andresuribe87, line 21: Can't tell if this was existing, but the tests could be separated 
```suggestion
func TestDIDDocumentMetadata_IsValid(t *testing.T) {
	t.Run("returns true with empty", func(t *testing.T) {
        	var metadata DocumentMetadata
                assert.True(t, metadata.IsValid())
	})


	t.Run("returns false when created field is not a timestamp", func(t *testing.T) {
	  badMetadata := DocumentMetadata{
		Created: "bad",
		Updated: time.Now().UTC().Format(time.RFC3339),
	  }
	  assert.False(t, badMetadata.IsValid())
	})
}
```

did/peer/peer2.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil, errors.Wrap(util.CastingError, "did:peer")
+	}
+
+	// The '=' at the end is an artifact of the encoding, and will mess up the decoding
+	// over the partials, so is removed.
+	// https://identity.foundation/peer-did-method-spec/index.html#generation-method
+	// ds := string(d)
+	// if ds[len(ds)-1] == '=' {
+	// 	d = DIDPeer(ds[:len(ds)-1])
+	// }
```
andresuribe87, line 41: Can this be deleted?

did/peer/peer_test.go
```diff
@@ -1,13 +1,27 @@
...
+func TestEncodePublicKeyWithKeyMultiCodecType(t *testing.T) {
+	// unsupported type
+	_, err := encodePublicKeyWithKeyMultiCodecType(crypto.KeyType("unsupported"), nil)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "not a supported key type")
+
+	// bad public key
+	_, err = encodePublicKeyWithKeyMultiCodecType(crypto.Ed25519, nil)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "unknown public key type; could not convert to bytes")
```
andresuribe87, line 22: Similar comment to the above. Though I see this was only moved. Up to you.

andresuribe87 (approved):
Thanks!

did/resolver/resolver.go
decentralgabe: good idea

did/peer/peer2.go
decentralgabe: I'm not familiar with the code so would prefer leaving it

did/peer/peer_test.go
decentralgabe: done

PR #385, opened 2023-05-15
Fix resolution metadata validation with datetime

andorsk:
The resolution validation code was failing due to the following: 

If you go to validator.v10 docs:

https://github.com/go-playground/validator/blob/c8b98cbcea08971fbead222ddf5fd406a40d40ed/validator_test.go#L12041

You will see that it requires to specify format of datetime string, otherwise, this test will fail: 

```go
t.Run("test valid time", func(t *testing.T) {
    now := time.Now().UTC().Format(time.RFC3339)
    var metadata = DocumentMetadata{
	Created: now,
    }
    assert.True(t, metadata.IsValid())
})
```

with the datetime validation failing. I have added a test to validate that the created field can properly be validate against the resolution metdata, and that the datetime field for validator is compared against RFC3339 specification (see below)

### DIFF
--- a/did/resolution/model.go
+++ b/did/resolution/model.go
@@ -24,8 +24,8 @@ func (r *ResolutionResult) IsEmpty() bool {
 
 // DocumentMetadata https://www.w3.org/TR/did-core/#did-document-metadata
 type DocumentMetadata struct {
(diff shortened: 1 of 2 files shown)


decentralgabe (approved):
nice

Issue #349, opened 2023-04-12
Support a local ION resolver for long form DIDs


decentralgabe:
based on https://github.com/TBD54566975/ssi-sdk/blob/main/did/ion/did.go#L46

could be added here https://github.com/TBD54566975/ssi-service/blob/main/internal/did/resolver.go#L33

decentralgabe:
fixed in #389

Issue #331, opened 2023-03-28
Utilize DID Resolution Options


decentralgabe:
Whenever I see `any`, it's typically a symptom of an API that hasn't been fleshed out yet. In this case, it's unclear the shape of what `ResolutionOption` is acceptable by the functions that use it. 

Instead, I would define an interface. Mind creating an issue to address?

Separately, This doesn't seem to be used anywhere.

_Originally posted by @andresuribe87 in https://github.com/TBD54566975/ssi-sdk/pull/329#discussion_r1150546075_

PR #389, opened 2023-05-22
Support ION DID Reconstruction & Long Form DID resolution

decentralgabe:
#349 

adds the ability to reconstruct any ION DID's state from a set of patches

### DIFF
--- a/did/ion/did.go
+++ b/did/ion/did.go
@@ -110,7 +110,7 @@ func LongToShortFormDID(longFormDID string) (string, error) {
 }
 
 // PatchesToDIDDocument applies a list of sidetree state patches in order resulting in a DID Document.
-func PatchesToDIDDocument(shortFormDID, longFormDID string, patches []any) (*did.Document, error) {
+func PatchesToDIDDocument(shortFormDID, longFormDID string, patches []Patch) (*did.Document, error) {
 	if len(patches) == 0 {
 		return nil, errors.New("no patches to apply")
 	}
@@ -123,16 +123,12 @@ func PatchesToDIDDocument(shortFormDID, longFormDID string, patches []any) (*did
 		AlsoKnownAs: longFormDID,
 	}
 	for _, patch := range patches {
-		knownPatch, err := tryCastPatch(patch)
-		if err != nil {
-			return nil, err
-		}
-		switch knownPatch.(type) {
-		case AddServicesAction:
-			addServicePatch := knownPatch.(AddServicesAction)
+		switch patch.GetAction() {
+		case AddServices:
+			addServicePatch := patch.(AddServicesAction)
 			doc.Services = append(doc.Services, addServicePatch.Services...)
-		case RemoveServicesAction:
-			removeServicePatch := knownPatch.(RemoveServicesAction)
+		case RemoveServices:
+			removeServicePatch := patch.(RemoveServicesAction)
 			for _, id := range removeServicePatch.IDs {
 				for i, service := range doc.Services {
 					if service.ID == id {
@@ -140,22 +136,22 @@ func PatchesToDIDDocument(shortFormDID, longFormDID string, patches []any) (*did
 					}
 				}
 			}
-		case AddPublicKeysAction:
-			addKeyPatch := knownPatch.(AddPublicKeysAction)
+		case AddPublicKeys:
+			addKeyPatch := patch.(AddPublicKeysAction)
 			gotDoc, err := addPublicKeysPatch(doc, addKeyPatch)
 			if err != nil {
 				return nil, err
 			}
 			doc = *gotDoc
-		case RemovePublicKeysAction:
-			removeKeyPatch := knownPatch.(RemovePublicKeysAction)
+		case RemovePublicKeys:
+			removeKeyPatch := patch.(RemovePublicKeysAction)
 			gotDoc, err := removePublicKeysPatch(doc, removeKeyPatch)
 			if err != nil {
 				return nil, err
 			}
 			doc = *gotDoc
-		case ReplaceAction:
-			replacePatch := knownPatch.(ReplaceAction)
+		case Replace:
+			replacePatch := patch.(ReplaceAction)
 			gotDoc, err := replaceActionPatch(doc, replacePatch)
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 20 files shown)


did/ion/did.go
```diff
@@ -41,6 +58,12 @@ func CreateLongFormDID(recoveryKey, updateKey jwx.PublicKeyJWK, document Documen
 	return strings.Join([]string{shortFormDID, encoded}, ":"), nil
 }
 
+// IsLongFormDID checks if a string is a long form DID URI
+func IsLongFormDID(maybeLongFormDID string) bool {
+	split := strings.Split(maybeLongFormDID, ":")
+	return len(split) == 4
```
andresuribe87, line 64:
```suggestion
	return strings.Count(maybeLongFormDID, ":") == 3
```

did/ion/did.go
```diff
@@ -85,3 +108,216 @@ func LongToShortFormDID(longFormDID string) (string, error) {
 	}
 	return shortFormDID, nil
 }
+
+// PatchesToDIDDocument applies a list of sidetree state patches in order resulting in a DID Document.
+func PatchesToDIDDocument(shortFormDID, longFormDID string, patches []any) (*did.Document, error) {
```
andresuribe87, line 113: I strongly discourage the use of `any` unless there is a strong reason to do so. Can you clarify why it's preferable to have patches be `[]any`? 

An alternative is to define an interface that all patches implement.

```go
type Patch interface{
  isPatch()
}

func (a AddServicesAction) isPatch() { }
```

did/ion/did.go
```diff
@@ -85,3 +108,216 @@ func LongToShortFormDID(longFormDID string) (string, error) {
...
+	}
+	for _, patch := range patches {
+		knownPatch, err := tryCastPatch(patch)
+		if err != nil {
+			return nil, err
+		}
+		switch knownPatch.(type) {
+		case AddServicesAction:
+			addServicePatch := knownPatch.(AddServicesAction)
+			doc.Services = append(doc.Services, addServicePatch.Services...)
```
andresuribe87, line 133: If you do this, then you don't need to re-cast within each `case`. You can use `typedPatch`
```suggestion
		switch typedPatch := knownPatch.(type) {
		case AddServicesAction:
			doc.Services = append(doc.Services, typedPatch.Services...)
```

did/ion/did.go
```diff
@@ -85,3 +108,216 @@ func LongToShortFormDID(longFormDID string) (string, error) {
...
+		default:
+			return nil, fmt.Errorf("unknown patch type: %T", patch)
+		}
+	}
+	return &doc, nil
+}
+
+// tryCastPatch attempts to cast a patch to a known patch type
+func tryCastPatch(patch any) (any, error) {
+	switch patch.(type) {
```
andresuribe87, line 173: Same comment as above. 

Mixing `map[string]any` and typed structs leads to too much freedom and code that is harder to maintain. Can this API be structured so we can have devs avoid shooting themselves in the foot?

did/ion/did.go
```diff
@@ -85,3 +108,216 @@ func LongToShortFormDID(longFormDID string) (string, error) {
...
+	}
+	return &doc, nil
+}
+
+func removePublicKeysPatch(doc did.Document, patch RemovePublicKeysAction) (*did.Document, error) {
+	for _, id := range patch.IDs {
+		removed := false
+		for i, key := range doc.VerificationMethod {
+			if key.ID == id {
+				doc.VerificationMethod = append(doc.VerificationMethod[:i], doc.VerificationMethod[i+1:]...)
```
andresuribe87, line 285: There is a lot of copying around within this function. Is this OK?

did/ion/did.go
```diff
@@ -85,3 +108,216 @@ func LongToShortFormDID(longFormDID string) (string, error) {
...
+					if auth == id {
+						doc.Authentication = append(doc.Authentication[:j], doc.Authentication[j+1:]...)
+					}
+				}
+				for j, auth := range doc.AssertionMethod {
+					if auth == id {
+						doc.AssertionMethod = append(doc.AssertionMethod[:j], doc.AssertionMethod[j+1:]...)
+					}
+				}
+				for j, auth := range doc.Authentication {
```
andresuribe87, line 300: Should this be `KeyAgreement`?

did/ion/did.go
```diff
@@ -85,3 +108,216 @@ func LongToShortFormDID(longFormDID string) (string, error) {
...
+					if auth == id {
+						doc.KeyAgreement = append(doc.KeyAgreement[:j], doc.KeyAgreement[j+1:]...)
+					}
+				}
+				for j, auth := range doc.Authentication {
```
andresuribe87, line 305: `CapabilityInvocation`?

did/ion/did.go
```diff
@@ -85,3 +108,216 @@ func LongToShortFormDID(longFormDID string) (string, error) {
...
+					if auth == id {
+						doc.CapabilityInvocation = append(doc.CapabilityInvocation[:j], doc.CapabilityInvocation[j+1:]...)
+					}
+				}
+				for j, auth := range doc.Authentication {
```
andresuribe87, line 310: `CapabilityDelegation`?

did/ion/did.go
```diff
@@ -85,3 +108,216 @@ func LongToShortFormDID(longFormDID string) (string, error) {
...
+	for _, id := range patch.IDs {
+		removed := false
+		for i, key := range doc.VerificationMethod {
+			if key.ID == id {
+				doc.VerificationMethod = append(doc.VerificationMethod[:i], doc.VerificationMethod[i+1:]...)
+				removed = true
+
+				// TODO(gabe): in the future handle the case where the value is not a simple ID
+				// remove from all other key lists
+				for j, auth := range doc.Authentication {
```
andresuribe87, line 290: Would it be possible to un-nest this for loop and the following ones?

did/ion/did_test.go
```diff
@@ -92,3 +100,151 @@ func TestGetShortFormDIDFromLongFormDID(t *testing.T) {
 
 	assert.Equal(t, shortFormDID, "did:ion:EiDyOQbbZAa3aiRzeCkV7LOx3SERjjH93EXoIM3UoN4oWg")
 }
+
+func TestPatchesToDIDDocument(t *testing.T) {
+	t.Run("Bad patch", func(tt *testing.T) {
+		doc, err := PatchesToDIDDocument("did:ion:test", "", []any{struct{ Bad string }{Bad: "bad"}})
+		assert.Empty(tt, doc)
+		assert.Error(tt, err)
+		assert.Contains(tt, err.Error(), "unknown patch type")
```
andresuribe87, line 109: Here and elsewhere
```suggestion
		assert.ErrorContains(tt, err, "unknown patch type")
```

did/ion/did.go
decentralgabe: will switch to interface

did/ion/did_test.go
decentralgabe: done

did/ion/did.go
decentralgabe: just a way to remove an element in an array should be fine

did/ion/did.go
decentralgabe: done

did/ion/model.go
```diff
@@ -101,18 +119,85 @@ type UpdateSignedDataObject struct {
...
+	for _, patch := range allPatches {
+		currPatch, ok := patch.(map[string]any)
+		if !ok {
+			return fmt.Errorf("patch is not a map")
+		}
+		action, ok := currPatch["action"]
+		if !ok {
+			return fmt.Errorf("patch has no action")
+		}
+		currPatchBytes, err := json.Marshal(currPatch)
```
andresuribe87, line 150: I think there is a way to use json.RawMessage to avoid the whole unmarshal -> marshal -> unmarshal round tripping, but I've never done it. If you're feeling brave :D

did/ion/did.go
decentralgabe: sorry thought I included this

did/ion/did.go
```diff
@@ -85,3 +108,154 @@ func LongToShortFormDID(longFormDID string) (string, error) {
...
+		for i, key := range doc.VerificationMethod {
+			if key.ID != id {
+				continue
+			}
+			doc.VerificationMethod = append(doc.VerificationMethod[:i], doc.VerificationMethod[i+1:]...)
+			removed = true
+
+			// TODO(gabe): in the future handle the case where the value is not a simple ID
+			// remove from all other key lists
+			for j, auth := range doc.Authentication {
```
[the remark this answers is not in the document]
decentralgabe: how? this needs to be nested as far as I can tell

did/ion/model.go
decentralgabe: hm you're right but I am not feeling brave

did/resolution/resolver.go
```diff
@@ -12,13 +12,13 @@ import (
 	"github.com/TBD54566975/ssi-sdk/did"
 )
 
-// ResolutionOption https://www.w3.org/TR/did-spec-registries/#did-resolution-options
-type ResolutionOption any
+// Option https://www.w3.org/TR/did-spec-registries/#did-resolution-options
+type Option any
```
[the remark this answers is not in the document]
decentralgabe: we can fix this when we implement any options
#331

did/ion/did.go
```diff
@@ -85,3 +108,154 @@ func LongToShortFormDID(longFormDID string) (string, error) {
...
+				if auth == id {
+					doc.Authentication = append(doc.Authentication[:j], doc.Authentication[j+1:]...)
+				}
+			}
+			for j, auth := range doc.AssertionMethod {
+				if auth == id {
+					doc.AssertionMethod = append(doc.AssertionMethod[:j], doc.AssertionMethod[j+1:]...)
+				}
+			}
+			for j, auth := range doc.Authentication {
```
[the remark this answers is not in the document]
decentralgabe: fixed

did/ion/did.go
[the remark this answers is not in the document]
andresuribe87: i have no idea what I was talking about... ignore please.

andresuribe87 (approved):

PR #403, opened 2023-05-31
Small refactors for cryptosuites; provable

### LINKED ISSUE #317: Consider Refactoring Provable
From https://github.com/TBD54566975/ssi-sdk/pull/288/files#r1115064918, we might want to rename provable. Possibly do more, since it's being used to signal that an object can be marhsalled into JSON, and that representation will have a "proof" object.

decentralgabe:
Fixes #317 - I think

- moved suites into their own directories
- added support for generic cred utility to handle the case where the credential is a json object


credential/util.go
```diff
@@ -33,8 +33,13 @@ func ToCredential(genericCred any) (jws.Headers, jwt.Token, *VerifiableCredentia
 		verifiableCredential := genericCred.(VerifiableCredential)
 		return nil, nil, &verifiableCredential, nil
 	case string:
-		// TODO(gabe) support the case where the string is a json representation of an LD credential
-		// JWT
+		// first try the case where the string is JSON of a VC object
+		var cred VerifiableCredential
+		if err := json.Unmarshal([]byte(genericCred.(string)), &cred); err == nil {
```
andresuribe87, line 38: instead of casting here, you can use the object from the swith

```go
switch typedCred := genericCred.(type) {
  // ...
  case string:
    var cred VerifiableCredential
    if err := json.Unmarshal([]byte(typedCred), &cred); err == nil {
    // ...
}
```

cryptosuite/jws2020/testutil_test.go
```diff
@@ -0,0 +1,32 @@
...
+	Issuer            any           `json:"issuer" validate:"required"`
+	IssuanceDate      string        `json:"issuanceDate" validate:"required"`
+	ExpirationDate    string        `json:"expirationDate,omitempty"`
+	CredentialStatus  any           `json:"credentialStatus,omitempty" validate:"omitempty,dive"`
+	CredentialSubject any           `json:"credentialSubject" validate:"required"`
+	CredentialSchema  any           `json:"credentialSchema,omitempty" validate:"omitempty,dive"`
+	RefreshService    any           `json:"refreshService,omitempty" validate:"omitempty,dive"`
+	TermsOfUse        []any         `json:"termsOfUse,omitempty" validate:"omitempty,dive"`
+	Evidence          []any         `json:"evidence,omitempty" validate:"omitempty,dive"`
+	Proof             *crypto.Proof `json:"proof,omitempty"`
```
andresuribe87, line 23: I think this small change would fail tests. If so, then let's keep the issue open. 
```suggestion
	Proof             *crypto.Proof
```

andresuribe87 (approved):

credential/util.go
decentralgabe: thanks updated here and elsewhere

cryptosuite/jws2020/testutil_test.go
decentralgabe: it would fail if there isn't a json property called `proof`

should I make an amending commit to rename the interface to `WithEmbeddedJSONProof`?

PR #409, opened 2023-06-09
Create release-notifier.yml

michaelneale:
This will notify devrel via ticket and also discord there is a new release to check against docs.

### DIFF
--- a/.github/workflows/release-notifier.yml
+++ b/.github/workflows/release-notifier.yml
@@ -1,0 +1,33 @@
+name: Create Issue On Release
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  create-issue:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Create Issue
+      uses: actions/github-script@v6
+      with:
+        github-token: ${{ secrets.DOCS_PAT }}  # Assuming 'PAT' is your Personal Access Token stored as a secret
+        script: |
+          const issueTitle = `A new release ${context.payload.release.tag_name} just cut for ${context.payload.repository.name}.`;
+          const issueBody = `Please check the api docs for compatibility with the release. Here is the [link to release notes](${context.payload.release.html_url}).`;
+          const labels = ['triage', 'priority-1-blocker', 'documentation'];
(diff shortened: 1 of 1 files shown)


.github/workflows/release-notifier.yml
```diff
@@ -0,0 +1,33 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Create Issue
+      uses: actions/github-script@v6
+      with:
+        github-token: ${{ secrets.DOCS_PAT }}  # Assuming 'PAT' is your Personal Access Token stored as a secret
+        script: |
+          const issueTitle = `A new release ${context.payload.release.tag_name} just cut for ${context.payload.repository.name}.`;
+          const issueBody = `Please check the api docs for compatibility with the release. Here is the [link to release notes](${context.payload.release.html_url}).`;
```
andresuribe87, line 18: Is `api docs` correct? Could we articulate in more detail?

andresuribe87 (approved):

.github/workflows/release-notifier.yml
michaelneale: it is really just a signal for devrel so doesn't matter too much what it says (just api docs are most serious - not too much an issue for ssi but has been for web5). It is the same signal the use for that to ensure things are ok.

.github/workflows/release-notifier.yml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  create-issue:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Create Issue
+      uses: actions/github-script@v6
+      with:
+        github-token: ${{ secrets.DOCS_PAT }}  # Assuming 'PAT' is your Personal Access Token stored as a secret
```
decentralgabe, line 15: whose PAT is this? do we have a way to manage them? have run into issues where they expire and we're not sure whose they were

michaelneale: this one won't expire. If it fails then can remove this action, or update it with a new one any time. Hopefully in future a github token will do the trick but haven't worked that out yet.

PR #412, opened 2023-06-09
Bye bye 3p jsonpath

andresuribe87:
Less deps = fun!

UPDATE: I didn't realize when I opened this PR that the jsonpath functionality was from `goccy/json`, and that it doesn't support [filter selector expressions](https://www.ietf.org/archive/id/draft-ietf-jsonpath-base-14.html#name-filter-selector). I'm unsure whether we should be supporting those complex expressions in input filters, but I'm putting this on hold in the meantime.


decentralgabe (approved):
nice

andresuribe87:
The test that's failing in question is https://github.com/TBD54566975/ssi-sdk/blob/effaa78729d6995182e56a6b06ab00f4aac94433/credential/exchange/submission_test.go#L693

Given that we need to support any json path expression, I'm closing this.

PR #433, opened 2023-07-25
Did Web Validation

andresuribe87:
The method implemented a very naive validation. Now, we return errors when the content-type isn't JSON, or when the body cannot be deserialized to a DID document.

### DIFF
--- a/did/web/web.go
+++ b/did/web/web.go
@@ -37,7 +37,7 @@ func (d DIDWeb) Validate(ctx context.Context) error {
 		return errors.Wrap(err, "resolving doc bytes")
 	}
 	contentType := header.Get("Content-Type")
-	if contentType != "application/json" {
+	if contentType != util.JSONContentType {
 		return errors.Errorf("header Content-Type received was `%s` but expected `%s`", contentType, "application/json")
 	}
 	var doc did.Document
--- a/did/web/web.go
+++ b/did/web/web.go
@@ -32,11 +32,19 @@ const (
 
 // Validate return nil if DID is valid, otherwise the validation error.
 func (d DIDWeb) Validate(ctx context.Context) error {
-	_, err := d.resolveDocBytes(ctx)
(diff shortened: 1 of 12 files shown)


did/web/web.go
```diff
@@ -32,11 +32,19 @@ const (
...
 // Validate return nil if DID is valid, otherwise the validation error.
 func (d DIDWeb) Validate(ctx context.Context) error {
-	_, err := d.resolveDocBytes(ctx)
+	docBytes, header, err := d.resolveDocBytes(ctx)
 	if err != nil {
 		return errors.Wrap(err, "resolving doc bytes")
 	}
-	return nil
+	contentType := header.Get("Content-Type")
+	if contentType != "application/json" {
```
decentralgabe, line 40: extract the content type to a const, consider adding it to a util file
assuming there's no stdlib const

did/web/web.go
```diff
@@ -157,26 +165,28 @@ func (d DIDWeb) Resolve(ctx context.Context) (*did.Document, error) {
...
-		return nil, errors.Wrapf(err, "constructing doc request %+v", docURL)
+		return nil, nil, errors.Wrapf(err, "constructing doc request %+v", docURL)
 	}
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
-		return nil, errors.Wrapf(err, "getting doc %+v", docURL)
+		return nil, nil, errors.Wrapf(err, "getting doc %+v", docURL)
 	}
-	defer resp.Body.Close()
+	defer func(Body io.ReadCloser) {
```
decentralgabe, line 184:
```suggestion
	defer func(body io.ReadCloser) {
```

decentralgabe (approved):
approving given changes

did/web/web.go
andresuribe87: Done.

did/web/web.go
andresuribe87: Done.

PR #437, opened 2023-07-31
Added docs and fixed required fields for resolution

andresuribe87:
title says it all


decentralgabe (approved):
good to add these properties, can we update any of the resolvers to make use of them?

andresuribe87:
> good to add these properties, can we update any of the resolvers to make use of them?

They're already being used in the `ion` resolver. Some fields don't make sense for other DID Methods. This PR is just adding documentation.

PR #444, opened 2023-08-08
handle kids in different forms; add local ion resolver

decentralgabe:


did/util_test.go
```diff
@@ -199,6 +199,32 @@ func TestGetKeyFromVerificationInformation(t *testing.T) {
...
+
+		key, err := GetKeyFromVerificationMethod(docWithHash, "test-did#test-kid")
+		assert.NoError(t, err)
+		assert.Equal(t, pubKey, key)
+
+		docWithHash.VerificationMethod[0].ID = "test-kid"
+		docWithoutHash := docWithHash
+		key, err = GetKeyFromVerificationMethod(docWithoutHash, "test-did#test-kid")
+		assert.NoError(t, err)
+		assert.Equal(t, pubKey, key)
```
andresuribe87, line 225: Separate this to a different sub-test for consistency.

andresuribe87, line 225: Separate this to a different sub-test for consistency.

did/util.go
```diff
@@ -62,12 +62,22 @@ func GetKeyFromVerificationMethod(did Document, kid string) (gocrypto.PublicKey,
 
 // matchesKIDConstruction checks if the targetID matches possible combinations of the did and kid
 func matchesKIDConstruction(did, kid, targetID string) bool {
-	maybeKID1 := kid                            // the kid == the kid
-	maybeKID2 := fmt.Sprintf("#%s", kid)        // the kid == the fragment with a #
-	maybeKID3 := fmt.Sprintf("%s#%s", did, kid) // the kid == the DID ID + the fragment with a #
-	maybeKID4 := fmt.Sprintf("%s%s", did, kid)  // the kid == the DID ID + the fragment without a #
-	maybeKID5, found := strings.CutPrefix(kid, did)
-	return targetID == maybeKID1 || targetID == maybeKID2 || targetID == maybeKID3 || targetID == maybeKID4 || (found && targetID == maybeKID5)
+	maybeKID1 := kid                                // the kid == the kid
```
andresuribe87, line 65: Cascading conditions make more sense now that there are 7 guesses. You can do that with a switch

andresuribe87 (approved):
Guessing what the intentions of the JWT signer are instead of making them explicit makes me think that this can be exploited in some way. I can't think of exactly how, but there's something off about this.

did/util_test.go
decentralgabe: done

did/util.go
decentralgabe: I tried that but found it much less readable...

PR #454, opened 2023-08-28
VC JSON Schema compliance with latest spec

decentralgabe:
- reworked schema validation logic to handle both credential and non credential types
- Added tests for all MUST values

### DIFF
--- a/credential/schema/vcjsonschema_test.go
+++ b/credential/schema/vcjsonschema_test.go
@@ -277,8 +277,7 @@ func TestIsCredentialValidForJSONSchema_JsonSchemaCredential(t *testing.T) {
 			schema["credentialSubject"] = schemaSubject
 
 			err := IsCredentialValidForJSONSchema(cred, schema, JSONSchemaCredentialType)
-			assert.Error(t, err)
-			assert.Contains(t, err.Error(), "not a valid URI")
+			assert.ErrorContains(t, err, "not a valid URI")
 		})
 	})
 
--- a/credential/schema/vcjsonschema_test.go
+++ b/credential/schema/vcjsonschema_test.go
@@ -22,7 +22,8 @@ var (
 	testVectors embed.FS
 )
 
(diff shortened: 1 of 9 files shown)


credential/schema/vcjsonschema_test.go
```diff
@@ -49,11 +49,351 @@ func TestValidateCredentialAgainstSchema(t *testing.T) {
...
+			invalidURI := "bad"
+			schemaSubject := schema["credentialSubject"].(map[string]any)
+			schemaSubjectSchema := schemaSubject["jsonSchema"].(map[string]any)
+			schemaSubjectSchema["$id"] = invalidURI
+			schemaSubject["jsonSchema"] = schemaSubjectSchema
+			schema["credentialSubject"] = schemaSubject
+
+			err := IsCredentialValidForJSONSchema(cred, schema, JSONSchemaCredentialType)
+			assert.Error(t, err)
+			assert.Contains(t, err.Error(), "not a valid URI")
```
andresuribe87, line 281: Here and elsewhere
```suggestion
			assert.ErrorContains(t, err, "not a valid URI")
```

credential/schema/vcjsonschema_test.go
```diff
@@ -49,11 +49,351 @@ func TestValidateCredentialAgainstSchema(t *testing.T) {
...
+				"properties": map[string]any{
+					"emailAddress": map[string]any{
+						"type": "string",
+					},
+				},
+				"required": []string{"emailAddress"},
+			},
+		},
+		"required": []string{"credentialSubject"},
+	}
```
andresuribe87, line 344: I would suggest putting the `json` and unmarshalling, instead of instantiating a map.

andresuribe87 (approved):
Couple nits.

credential/schema/vcjsonschema_test.go
decentralgabe: going to stick with a map since it's easier to maintain IMO

PR #455, opened 2023-08-29
JSON Schema - include latest spec updates around CredSchema/digestSRI

decentralgabe:
- Add enforcement of credential schema for json schema credentials and associated tests

### DIFF
--- a/credential/schema/vcjsonschema_test.go
+++ b/credential/schema/vcjsonschema_test.go
@@ -254,7 +254,7 @@ func TestIsCredentialValidForJSONSchema_JsonSchemaCredential(t *testing.T) {
 		})
 	})
 
-	t.Run("2.2 The value of the credentialSchema property MUST always be set to [known json schema]", func(t *testing.T) {
+	t.Run("The value of the credentialSchema property MUST always be set to [known json schema]", func(t *testing.T) {
 		t.Run("valid credentialSchema", func(t *testing.T) {
 			cred := getTestVCJSONSchemaCredential()
 			schema := getTestVCJSONSchemaSchema()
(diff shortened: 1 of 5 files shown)


credential/schema/vcjsonschema_test.go
```diff
@@ -254,6 +254,53 @@ func TestIsCredentialValidForJSONSchema_JsonSchemaCredential(t *testing.T) {
 		})
 	})
 
+	t.Run("2.2 The value of the credentialSchema property MUST always be set to [known json schema]", func(t *testing.T) {
```
andresuribe87, line 257: Can you clarify that the 2.2 refers to a specific test from teh spec test suite?

credential/schema/vcjsonschema_test.go
```diff
@@ -254,6 +254,53 @@ func TestIsCredentialValidForJSONSchema_JsonSchemaCredential(t *testing.T) {
...
+				"id":        JSONSchemaCredentialSchemaID,
+				"type":      "NotJsonSchema",
+				"digestSRI": JSONSchemaCredentialDigestSRI,
+			}
+			err := IsCredentialValidForJSONSchema(cred, schema, JSONSchemaCredentialType)
+			assert.Error(t, err)
+			assert.ErrorContains(t, err, "credential schema's credential schema type<NotJsonSchema> does not match known type<JsonSchema>")
+		})
+
+		t.Run("credentialSchema missing digestSRI", func(t *testing.T) {
```
andresuribe87, line 291: Can you add a test where the `digestSRI`, when computed from the source, doesn't match the one that was given in the VC?

decentralgabe: yes

credential/schema/vcjsonschema_test.go
decentralgabe: yes

andresuribe87 (commented):
We will also need to validate that a VC that's setting the `credentialSchema.digestSRI` property, has a value which matches against the locally calculated digestSRI. This typically involves dereferencing `credentialSchema.id`, and then following the spec of SubResource Integrity. 

I would prefer that we do that in this PR, so that the same mechanism is used for every VC (instead of special casing VCs who's `type` property is `JsonSchemaCredential`). That said, I'd understand leaving that for another PR. Up to you.

decentralgabe:
I'd rather take that in a separate PR, but opened an issue https://github.com/TBD54566975/ssi-sdk/issues/456

andresuribe87 (approved):

PR #457, opened 2023-08-29
Go Mobile support for DID Key

decentralgabe:


decentralgabe:
CodeQL will be broken until the authors update it to support go1.21, should not block merge

mobile/did_key.go
andresuribe87, line 1: The methods in here aren't tested. I think we should have some sample code that exercises it. It can act as documentation as well on how it should be called.

magefile.go
```diff
@@ -305,3 +306,34 @@ func Vuln() error {
...
 	return installIfNotPresent("govulncheck", "golang.org/x/vuln/cmd/govulncheck@latest")
 }
+
+func installGoMobileIfNotPresent() error {
+	return installIfNotPresent(gomobile, "golang.org/x/mobile/cmd/gomobile@latest")
+}
+
+// IOS Generates the iOS packages
+// Note: this command also installs "gomobile" if not present
+func IOS() error {
```
andresuribe87, line 316: Should this target, as well as the Android target, be documented in a README?

crypto/keys.go
```diff
@@ -67,7 +66,8 @@ type Option struct {
 }
 
 var (
-	ECDSAMarshalCompressed = Option{Name: "ecdsa-compressed", Value: true}
+	ECDSAMarshalCompressed   = Option{Name: "ecdsa-compressed", Value: true}
+	ECDSAUnmarshalCompressed = Option{Name: "ecdsa-compressed", Value: true}
```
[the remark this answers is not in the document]
decentralgabe: yes, one used for marshaling one for unmarshaling

I could condense to a `ECDSACompressed` option

mobile/did_key.go
decentralgabe: agree will add tests

magefile.go
decentralgabe: yes good call

crypto/keys.go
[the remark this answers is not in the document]
andresuribe87: I generally discourage using an `Option` type which has a field named `Name`. It's bypasses the type system, and makes it easy to shoot yourself in the foot.

I would suggest using an enum es below: 
```go
const (
  ECDSAMarshalCompressed int = iota
  ECDSAUnmarshalCompressed
)
```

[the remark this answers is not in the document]
decentralgabe: done

decentralgabe:
tests added, comments made, bugs fixed. should be good to go

mobile/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+## iOS
+
+```
+mage ios
+```
+
+## Android
+
+```
```
andresuribe87, line 15: I think it's important to cover:
* Documentation on how to include this in an android/ios project, with a simple example. 
* CI that exercises actual Android/ios code so we can be confident this works.
* Guidance to readers that tells them how to discover available functionality within the SDK. 

Of course, that's a lot of scope for this PR, so tracking in separate issues is cool. But I think this is critically important.

mobile/did_key.go
```diff
@@ -0,0 +1,140 @@
...
+	for _, kt := range key.GetSupportedDIDKeyTypes() {
+		keyTypes = append(keyTypes, string(kt))
+	}
+	return keyTypes
+}
+
+type GenerateDIDKeyResult struct {
+	DID string         `json:"did"`
+	JWK map[string]any `json:"jwk"`
+}
```
andresuribe87, line 23: godocs for all fields here. This should be treated this as a public API

mobile/did_key.go
```diff
@@ -0,0 +1,140 @@
...
+
+// GenerateDIDKey generates a new DID key and returns a JSON representation of GenerateDIDKeyResult
+func GenerateDIDKey(kt string) ([]byte, error) {
```
andresuribe87, line 26: Can you include the possible kt values in the godoc?

mobile/did_key.go
```diff
@@ -0,0 +1,140 @@
...
+		DID: didKey.String(),
+		JWK: jwk,
+	}
+	return json.Marshal(result)
+}
+
+type CreateDIDKeyRequest struct {
+	KeyType      string         `json:"keyType"`
+	PublicKeyJWK map[string]any `json:"publicKeyJwk"`
+}
```
andresuribe87, line 67: godocs for all types here.

mobile/did_key.go
```diff
@@ -0,0 +1,140 @@
...
+
+type CreateDIDKeyResult struct {
+	DID string `json:"did"`
+}
```
andresuribe87, line 71: godocs, hopefully with an example. Something like 
```
// DID that was created. E.g. `did:key:zHjkdjd...`
```

mobile/did_key.go
```diff
@@ -0,0 +1,140 @@
...
+		logrus.WithError(err).Error("failed to create did key")
+		return nil, err
+	}
+
+	result := CreateDIDKeyResult{DID: didKey.String()}
+	return json.Marshal(result)
+}
+
+type Document struct {
+	DIDDocument map[string]any `json:"didDocument"`
```
andresuribe87, line 115: You know what I want :)

mobile/did_key.go
```diff
@@ -0,0 +1,140 @@
...
+	PublicKeyJWK map[string]any `json:"publicKeyJwk"`
+}
+
+type CreateDIDKeyResult struct {
+	DID string `json:"did"`
+}
+
+// CreateDIDKey creates a new DID key from an existing public key, accepting a JSON representation of CreateDIDKeyRequest
+// and returns a JSON representation of CreateDIDKeyResult which contains the DID of the newly created key as a string
+func CreateDIDKey(requestBytes []byte) ([]byte, error) {
```
andresuribe87, line 75: Let's make it easy for clients to understand what they should put into `requestBytes`: can we type this?

andresuribe87 (approved):
Non-blocking comments.

mobile/did_key.go
decentralgabe: I'll add that the values are the result of the GetSupportedDIDKeyTypes

mobile/did_key.go
andresuribe87: It's tough to say, because I don't know how a mobile dev would interact with this. I'm of the opinion that a worthy goal is to have mobile devs not have to read any go code at all.

So I ask myself, as a mobile dev, do I know what I should put in here? I think the current answer is probably not. I would make a suggestion if I understood better how calling from Java/Kotlin, Swift would look like.

decentralgabe: I think go docs are the best we can realistically do...if they don't read the API docs they'll always be lost

andresuribe87: How about having the type of the parameter this function accepts be `CreateDIDKeyRequest`? According to the [documentation](https://pkg.go.dev/golang.org/x/mobile/cmd/gobind#hdr-Type_restrictions), structs are allowed.

It feels like it's a bit easier to deal with a typed object in Kotlin / Swift.

I was hoping that documentation gets automatically translated into the artifact that's created (the AAR or the Obj-C lib), but it unfortunately doesn't look like that's the case. That's unfortunate :( 
To mitigate, I would say adding an example in the target language that demonstrates how to call this function would be great.

mobile/README.md
decentralgabe: https://github.com/TBD54566975/ssi-sdk/issues/459

mobile/did_key.go
decentralgabe: will take this in a follow up

PR #458, opened 2023-08-30
Refactoring so error causality is easier to follow

andresuribe87:
this leads to much better readability of error chains


decentralgabe (approved):
thanks for this

PR #488, opened 2023-11-21
set default key id as jwk thumbprint

decentralgabe:


decentralgabe:
@andresuribe87 keep in line with the other sdks

Issue #400, opened 2023-05-30
Update VerifiableCredentialBuilder


andresuribe87:
The [spec](https://www.w3.org/TR/vc-data-model-2.0/#identifiers) states that `id` is optional. If present, it must be a URL, which may be dereferenced.

To better support the spec, we should update the sdk such that:
* The default ID is empty (to foster stronger privacy).
* Update the method `SetID()`, so it does validation that the parameter is a URI (using the `url.Parse` function).

decentralgabe:
```Example id values include UUIDs (urn:uuid:0c07c1ce-57cb-41af-bef2-1b932b986873), HTTP URLs (https://id.example/things#123), and DIDs (did:example:1234abcd).```

since we use UUIDs we'll probably want to update to add the `urn:uuid` prefix

m-rit:
@decentralgabe, can I work on a PR to to add this fix if noone is working on it? I've used this repo for a personal projects  and I'd like to contribute to the issues.

decentralgabe:
@m-rit yes definitely

most of the work at TBD has moved to https://github.com/TBD54566975/web5-go, though I still use this SDK sometimes

m-rit:
I have raised a PR to address the issue,  NewVerifiableCredentialBuilder now creates VC with empty ID. Should I add an option argument in the function to aupport VC creation with new uuid (the original behaviour) ?

decentralgabe:
@m-rit yes!

PR #532, opened 2024-06-05
Update ID rules in NewVerifiableCredentialBuilder

m-rit:
Address Issue #400 .
 Changes default id in NewVerifiableCredentialBuilder as empty, validates if id is a parsable uri in SetID

### DIFF
--- a/credential/builder.go
+++ b/credential/builder.go
@@ -32,7 +32,7 @@ type VerifiableCredentialBuilder struct {
 }
 
 // NewVerifiableCredentialBuilder returns an initialized credential builder with some default fields populated
-// Default id is empty
+// Setting emptyID as True creates VC with empty ID. Default ID is a random UUID.
 func NewVerifiableCredentialBuilder(emptyID ...bool) VerifiableCredentialBuilder {
 	contexts := []string{VerifiableCredentialsLinkedDataContext}
 	types := []string{VerifiableCredentialType}
--- a/credential/builder.go
+++ b/credential/builder.go
@@ -33,19 +33,25 @@ type VerifiableCredentialBuilder struct {
 
 // NewVerifiableCredentialBuilder returns an initialized credential builder with some default fields populated
 // Default id is empty
-func NewVerifiableCredentialBuilder() VerifiableCredentialBuilder {
+func NewVerifiableCredentialBuilder(emptyID ...bool) VerifiableCredentialBuilder {
 	contexts := []string{VerifiableCredentialsLinkedDataContext}
 	types := []string{VerifiableCredentialType}
-	return VerifiableCredentialBuilder{
+	id := uuid.NewString()
+	if len(emptyID)>0 && emptyID[0]{
+		id = ""
+	}
+	vcb := VerifiableCredentialBuilder{
 		contexts: contexts,
 		types:    types,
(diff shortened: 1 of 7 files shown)


credential/builder.go
```diff
@@ -31,14 +32,15 @@ type VerifiableCredentialBuilder struct {
...
+// Default id is empty
 func NewVerifiableCredentialBuilder() VerifiableCredentialBuilder {
 	contexts := []string{VerifiableCredentialsLinkedDataContext}
 	types := []string{VerifiableCredentialType}
 	return VerifiableCredentialBuilder{
 		contexts: contexts,
 		types:    types,
 		VerifiableCredential: &VerifiableCredential{
-			ID:           uuid.NewString(),
+			ID:           "",
```
decentralgabe, line 43: 🤔 given that this is a required property I'm not sure it's a good idea to leave this empty

perhaps it should be added as an optional property to the builder, and if that is empty we generate it automatically?

m-rit: yeah, it's better to keep an option argument. I'll make the changes, thanks!

m-rit: Hi @decentralgabe, I have added an optional argument to keep ID empty. Kindly review.

credential/builder.go
```diff
@@ -33,19 +33,25 @@ type VerifiableCredentialBuilder struct {
 
 // NewVerifiableCredentialBuilder returns an initialized credential builder with some default fields populated
 // Default id is empty
-func NewVerifiableCredentialBuilder() VerifiableCredentialBuilder {
+func NewVerifiableCredentialBuilder(emptyID ...bool) VerifiableCredentialBuilder {
```
decentralgabe, line 36: what do you think about making this a bit more readabile?

we could change the build to accept a new type like `IDvalue` for example


```go

type IDValue string

const EmptyIDValue IDValue = ""

func NewVerifiableCredentialBuilder(id IDValue) VerifiableCredentialBuilder {
    if id == EmptyIDValue {
        // generate ID
    }
    ...
}
```

m-rit: Hi @decentralgabe, thanks for your inputs. 
Please confirm if the behaviour is as follows: 
```
NewVerifiableCredentialBuilder (EmptyIDValue)  -> creates vc with id as uuid
NewVerifiableCredentialBuilder  ("xxx") -> generates vc with id as "xxx"
vcb.SetID("") ->  sets ID as empty
```

decentralgabe: ah, good consideration.

Perhaps we should add....

`EmptyIDValue` -> set ID value to empty
`GenerateIDValue` -> generate a uuid
`CustomIDValue("<your-id-here>")` -> use custom ID

to be unambiguous?

m-rit: Hi @decentralgabe , I have added IDValue argument here to all three cases (empty ID/ generate ID/ custom ID)

decentralgabe (approved):

decentralgabe:
thanks @m-rit I approved - you may have to take a look at the build issues

m-rit:
Hi @decentralgabe,  I have fixed all the build issues (the function signature for `NewVerifiableCredentialBuilder` had to be updated in all files). I think additional approval is needed for the workflows.

decentralgabe:
thanks @m-rit approved and merging

